PR #21, opened 2016-05-25
Travis CI Builds

samtstern:
@kroikie please review, and if you choose to merge please use the Gthub 'squash' option.

This does the following:

  * Standardize build tools version, ignoring lint, and proguarding release variant
  * Add Travis CI configuration:
    * On pull requests, do a quick assembleDebug for all folders (~10 min on Travis)
    * On merged commits, build the whole thing (~30 min on Travis)

Future work:

  * Make the build faster so that we don't ever hit Travis' 50 minute limit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,18 @@
+language: android
+android:
+  components:
+    - platform-tools
+    - tools
+    - build-tools-23.0.3
+    - android-23
+    - extra-google-google_play_services
+    - extra-google-m2repository
+    - extra-android-m2repository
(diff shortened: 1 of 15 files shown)


kroikie:
@samtstern could we travis encrypt the google-services.json file?

samtstern:
It's all just mock values I made up by smashing the keyboard.  The
important thing is they pass basic regex and match the package name so the
build can succeed.

kroikie:
Ah so no real API keys, gotcha!

PR #57, opened 2016-06-20
Update to android-24

friederbluemle:
This updates Gradle wrapper to `2.14` and all Android dependencies to `24`.


samtstern:
Appreciated as always!

PR #65, opened 2016-06-27
Fix App Indexing build

samtstern:
Change-Id: I107c73eb741a686847eedb63bf854b92bae3b4e9

### DIFF
--- a/app-indexing/app/build.gradle
+++ b/app-indexing/app/build.gradle
@@ -13,11 +13,16 @@ android {
 
         testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
     }
+
     buildTypes {
         release {
             minifyEnabled true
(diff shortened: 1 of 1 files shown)


friederbluemle:
@samtstern What was the Lint error you were seeing? The build passes with only warning for me locally (no need to ignore errors).

friederbluemle:
All projects now build successfully without ignoring errors: #66

samtstern:
@friederbluemle the strange thing is the lint error was only occurring on Travis CI, and since app-indexing was the only sub-project not already ignoring them I just added that to get the build back to green.

Thanks a ton for #66, I will take a look now.

friederbluemle:
@samtstern Yeah, I just noticed the Travis CI error after submitting #66. I think the error is because of an outdated `platform-tools` component on Travis. I added commit cdb0224d52fa25bd9f23ca08f403dbf371194f27 which should fix it.

PR #66, opened 2016-06-29
Fix Lint errors

friederbluemle:
All projects now build successfully without ignoring Lint errors (which is almost never a good idea without a strong reason).

There were only a couple of Lint errors. I fixed them in two separate commits:
The first one in the `analytics` project related to a resource style value `colorAccent` being used on API < 21. The other error was in the `auth` project related to Facebook SDK dependencies being referenced in the `main` source set. Since the project's **app/build.gradle** specifies the dependency as `facebookCompile` they are **only** available in the `facebook` variant of the project, thus I moved the declarations to a new **AndroidManifest.xml** file in the respective folder.


samtstern:
@friederbluemle thanks for taking this on!  Since the Travis build passes I am going to merge them in.  Very much appreciated, I was certainly being lazy by ignoring lint errors.

PR #90, opened 2016-07-18
Handle both notification and data payloads

kroikie:
Update to check for and handle both data and notification messages in onMessageReceived callback.


samtstern:
@kroikie travis is failing to find build tools 24.0.0 for some reason

Can you do the following locally:
```
TRAVIS_PULL_REQUEST=1 ./build.sh
```

samtstern:
Also bonus points if you figure out why Travis is failing and fix it at the same time.

kroikie:
Builds successfully locally, so there seems to be an issue with travis getting the build tools. Investigating.

PR #99, opened 2016-08-02
Update Gradle wrapper to 2.14.1

friederbluemle:


samtstern:
@friederbluemle Travis doesn't seem to like this one, can you look into the build failure?

friederbluemle:
@samtstern Interesting.. Sure, I'll take a look!

friederbluemle:
Okay, seems like the error was unrelated to the Gradle wrapper update.

    Could not GET 'https://maven.fabric.io/public/com/twitter/sdk/android/twitter/1.13.1/twitter-1.13.1.pom'. Received status code 503 from server: Service Unavailable: Back-end server is at capacity

I restarted the build using `git commit --amend --no-edit` and a force-push, and the build is passing now.

samtstern:
LGTM. Thanks for digging into that!

PR #101, opened 2016-08-05
Replace switches with if/else

samtstern:


samtstern:
@kroikie please review

PR #106, opened 2016-08-13
Change to minSDK version from 9 to 16

Dagdagger:
Min SDK version 9 caused a minSDK gradle build error when trying to use the FireBaseAuth package


Vivek-Tomar:
@Dagdagger **what about devices having api below 16 ?** FCM is intended to use with api level 9 and above.

samtstern:
@Dagdagger Firebase supports API level 9 and up, so this is not a desirable change.  If you are having an issue building this quickstart, please open a new issue.

PR #122, opened 2016-09-06
Handle both notification and data payloads

abshar-rafique:


samtstern:
I believe this pull request was made in error, it looks like this change is already in the master branch:
https://github.com/firebase/quickstart-android/blob/master/messaging/app/src/main/java/com/google/firebase/quickstart/fcm/MyFirebaseMessagingService.java#L58-L65

PR #132, opened 2016-09-14
Refactor Storage sample to download in background

samtstern:
Move download to a background service which notifies the user via a Notification when the download is in progress/completed.

This prevents bad UX if the user leaves the app before download is finished.


samtstern:
@kroikie please review.  Sorry, this is a big one.

kroikie:
Np I'll have a look.

samtstern:
Note to self: this will break includecode, do not submit without pending CL to fix that

kroikie:
LGTM, assuming the TODOs are addressed in a future PR.

samtstern:
Good call on the TODOs, I actually did address them but forgot to remove the notes.  Will update.

PR #135, opened 2016-09-17
correct a minor spelling mistake

l3d00m:


l3d00m:
I signed it!

samtstern:
@l3d00m thanks for the PR!  I don't think this is an actual mistake, @kroikie probably meant for "IID" to be shorthand for "Instance ID".

PR #143, opened 2016-09-24
Add FirebaseUI activity to Auth sample

samtstern:
Adding a FirebaseUI activity to the auth quickstart.

Also changing the build variants to be more accurate about what they are for (minSDK)


samtstern:
@kroikie PTAL next week :-)

PR #146, opened 2016-09-25
Can display push notification now

HasanFaraaz:
Added line "sendNotification(remoteMessage.getNotification().getBody());" which will now show the notificaiton in the status bar


HasanFaraaz:
I signed it!

kroikie:
Thanks for the PR however the sendNotification is an optional call, so would be best for us to leave it as a todo for the developer as the comment below your change indicates.

PR #149, opened 2016-09-27
Update Gradle wrapper to 3.1

friederbluemle:
This updates Gradle wrapper to `3.1`, and also increases the maximum Gradle daemon heap space which allows dex to run in process, speeding up the build.


samtstern:
Thanks!

PR #153, opened 2016-10-12
Make travis much faster

samtstern:
Two improvements:
  * Build matrix (thanks @iulukaya !)
  * Gradle caching

This gets us down from one big 30-minute build to individual 9-minute builds.

This also opens the door for actually running the tests on Travis!  But I'm not sure if our tests are sufficiently de-flaked for that.


samtstern:
screen shot 2016-10-12 at 2 14 01 pm

PR #216, opened 2017-01-24
Showcasing the Streaming download API to indicate progress

schmidt-sebastian:
We don't have an example of how to use the internals of the Streaming API. I added code to the upload and download handlers to exemplify how a developer could use the process data that we provide.

### DIFF
--- a/storage/app/src/main/java/com/google/firebase/quickstart/firebasestorage/MyBaseTaskService.java
+++ b/storage/app/src/main/java/com/google/firebase/quickstart/firebasestorage/MyBaseTaskService.java
@@ -43,7 +43,7 @@ public abstract class MyBaseTaskService extends Service {
      * Show notification with a progress bar.
      */
     protected void showProgressNotification(String caption, long completedUnits, long totalUnits) {
-        int percentComplete = 50;
+        int percentComplete = 0;
         if (totalUnits > 0) {
             percentComplete = (int) (100 * completedUnits / totalUnits);
         }
--- a/storage/app/src/main/java/com/google/firebase/quickstart/firebasestorage/MyBaseTaskService.java
+++ b/storage/app/src/main/java/com/google/firebase/quickstart/firebasestorage/MyBaseTaskService.java
@@ -1,6 +1,11 @@
 package com.google.firebase.quickstart.firebasestorage;
 
+import android.app.NotificationManager;
+import android.app.PendingIntent;
 import android.app.Service;
+import android.content.Context;
+import android.content.Intent;
+import android.support.v4.app.NotificationCompat;
 import android.util.Log;
 
 /**
@@ -8,6 +13,7 @@ import android.util.Log;
  * count is zero.
  */
 public abstract class MyBaseTaskService extends Service {
+    static final int NOTIFY_ID = 0;
 
     private static final String TAG = "MyBaseTaskService";
     private int mNumTasks = 0;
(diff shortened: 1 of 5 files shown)


samtstern:
Ok I just downloaded this and ran it:

  * On uploads the progress does not seem to work at all.  The dialog sits at 0 and then suddenly disappears when the upload is finished.
  * The download progress works nicely but the starting state says `NaN` which is not ideal (and I didn't even know Java had `NaN`)

samtstern:
Oh and another issue is what happens to the dialog when you rotate the app.  That's a minor thing since the notification stays in place.

schmidt-sebastian:
- The upload progress event only fires every 256KB, so for small files it does not fire at all. That's why you don't see any progress happen there.
- The download progress updates every 256 bytes, but it does get initialized with NaN since I didn't want to convolute the code with special casing.

After our discussion, I updated the app to only use the notification bar for progress updates to work around these issues. This makes smaller uploads feel smoother (the sudden jumps in progress happen out of sight), but in my point it's not as good for larger files since progress isn't immediately apparent anymore. This might be ok though since our goal is to showcase the API and not make polished UIs.

There are two remaining issues that I would like your help with (since I don't really know Android):
- For some reason, the final notification for the finished upload always shows 0 bytes, even though the value is set as part of the Intent.
- The switch to "Download finished" only works 5% of the time, while "Upload finished" shows up 100% of the time. I believe that the order of the notifications is correct.

samtstern:
Update looks great, two small issues left:

  1. The success progressdialog seems to always say "0 bytes downloaded" when successful
  2. After a successful download the notification gets stuck like this:

storage-notif

samtstern:
@schmidt-sebastian oops I see you were already aware of these issues, didn't see your comment.  I think this UI is fine for a sample minus those two lingering issue.  Using the notifications area in a simple way like this is a good balance between small code and something that would actually work in production.

samtstern:
@schmidt-sebastian I pushed a commit to fix those two issues.

  * "0 bytes" was caused by passing the extra as an int but retrieving it as a long
  * I think the notification not going away was caused by a race condition, seems to be solved by using a different notification for progress and finished and explicitly dismissing the progress one

I also moved a bunch of strings to `strings.xml` to be more idiomatic.

storage/app/src/main/java/com/google/firebase/quickstart/firebasestorage/MyBaseTaskService.java
```diff
@@ -40,19 +42,24 @@ private synchronized void changeNumberOfTasks(int delta) {
     /**
      * Show notification with a progress bar.
      */
-    protected void showProgressNotification(String caption, int completedUnits, int totalUnits) {
+    protected void showProgressNotification(String caption, long completedUnits, long totalUnits) {
+        int percentComplete = 50;
```
schmidt-sebastian: Does this cause the progress to drop from 50% to 0% at the very beginning of the download?

schmidt-sebastian:
Thanks for your help, one small comment on your fix.

storage/app/src/main/java/com/google/firebase/quickstart/firebasestorage/MyBaseTaskService.java
samtstern: Oops good catch.  That was left over from debugging.

PR #232, opened 2017-03-01
Updated documentation regarding setting up sample apps

hiranya911:
Added instructions on adding sample apps to a Firebase project.


samtstern:
@hiranya911 thanks again for the helpful documentation update!

PR #237, opened 2017-03-03
Update Gradle wrapper and Android plugin

friederbluemle:
The stable version of Android Studio 2.3.0 was released.
https://android-developers.googleblog.com/2017/03/android-studio-2-3.html

This updates:
- Gradle wrapper to `3.4`
- Android Gradle plugin to `2.3.0`

PR #238, opened 2017-03-04
Update Gradle 

ercanpinar:
Update Gradle and 
Update Gradle wrapper


samtstern:
Sorry looks like #237 beat you to it!

PR #286, opened 2017-06-16
Speed up builds

samtstern:


samtstern:
Well the build is not any faster... but green!

PR #289, opened 2017-06-19
Need for build speed!

### LINKED ISSUE #288: Speed up Travis CI build
Right now our builds are very slow.  Each module takes about ~20m, even running them in parallel on Travis we get a total build time of close to an hour:

screenshot from 2017-06-19 08 58 46

I recently tried increasing the resources in #286 but it didn't have much of an effect.

The build for [FirebaseUI-Android](https://github.com/firebase/FirebaseUI-Android) is only about 7 minutes even though it's a multi-module build including test, which should be much slower than our build in theory.

(cc: @SUPERCILEX any ideas?)

SUPERCILEX:
@samtstern at a high level overview, this PR does a few things:
- Updates the Android Gradle Plugin to 3.0 so we can take advantage of the improvements in multi-module project build speeds
- Similarly, update Gradle to 4.0
- Enable the new Gradle build cache which ensures only the modules that changed are rebuilt
- Require sudo for Travis which brings massive internet speed boosts
- Require a soul-crushing amount of memory to build the full multi-module project. (Which I'm hoping won't be an issue)

fixes #288


build.sh
```diff
@@ -32,12 +33,9 @@ ${ANDROID_HOME}/tools/bin/sdkmanager --channel=3 \
 # Build
 if [ $TRAVIS_PULL_REQUEST = false ] ; then
   # For a merged commit, build all configurations.
-  GRADLE_OPTS=$OPTS ./gradlew clean build
+  ./gradlew clean build
 else
   # On a pull request, just build debug which is much faster and catches
   # obvious errors.
-  GRADLE_OPTS=$OPTS ./gradlew clean :app:assembleDebug
+  ./gradlew clean assembleDebug
```
SUPERCILEX: The other important thing we do that I forgot to mention explicitly is that all modules are built in single Travis build. Since they don't depend on each other, we can build them in parallel without taking a performance hit which is faster than booting up a bunch of Travis VMs and redownloading stuff.

invites/app/build.gradle
```diff
@@ -21,32 +21,24 @@ android {
         }
     }
 
-    productFlavors {
-        main {
-            // TODO(developer): Replace this with your app code
-            //                  See: https://firebase.google.com/docs/dynamic-links/android
-            resValue "string", "app_code", "YOUR_APP_CODE"
```
SUPERCILEX: TODO(me): This is only used in a test, figure out why.

SUPERCILEX:
@samtstern Aw man, there's a bug when building stuff in parallel so we're up to ~6m build time now. 😢

admob/app/build.gradle
```diff
@@ -31,16 +31,15 @@ configurations.all {
...
 dependencies {
-    compile fileTree(dir: 'libs', include: ['*.jar'])
-    // AppCompat
-    compile 'com.android.support:appcompat-v7:25.3.1'
+    implementation fileTree(dir: 'libs', include: ['*.jar'])
+    implementation 'com.android.support:appcompat-v7:25.3.1'
 
     // [START gradle_play_config]
-    compile 'com.google.firebase:firebase-ads:11.0.1'
+    implementation 'com.google.firebase:firebase-ads:11.0.1'
```
samtstern: I'm fine to move to the 3.0 plugin but could we move these back to `compile` throughout?  Many developers using this project as a reference are not on the 3.0 tools yet and `implementation` will not work for them if they copy and paste (which everyone does).

samtstern:
@SUPERCILEX wow thank you for doing all this!  That's some serious speed.  Left one comment.

admob/app/build.gradle
SUPERCILEX: K, sounds reasonable. Do you want that for each module or just the admob one?

invites/app/build.gradle
samtstern: We need to keep this.  These samples are actually cloned and built by another team at Google and used for end-to-end Firebase testing and they requested we use this `app_code` in gradle config to make their lives easier.

SUPERCILEX: Ah crud, that's makes life harder because we need to add a dimension. Let's see if I can get it to work! 😄

SUPERCILEX:
@samtstern Bringing along the improvements from FirebaseUI, I also added compiling tests to the `check` task because I didn't noticed that I broke the tests by removing that Gradle string until I tried running them.

Speaking of tests, everything compiles locally and all the tests pass except for an auth one. I'm pretty that's not my fault, but could you try running `./gradlew connectedCheck` with your local head? (Not this PR) Thanks!

samtstern:
@SUPERCILEX I am pretty sure some of the tests are broken, you're right.  So let's leave them out of this PR for the sake of greenness!

SUPERCILEX:
@samtstern K, SGTM.

admob/app/build.gradle
samtstern: Need it for all modules

SUPERCILEX:
@samtstern I took one last look at everything and I believe https://github.com/firebase/quickstart-android/pull/289#discussion_r122856076 is the only thing left to resolve.

SUPERCILEX:
Oh, just saw your comment. I'm on it.

SUPERCILEX:
@samtstern Cool beans, I think I reverted everything correctly! (I need to file a bug with Tor to provide some sort of Android Studio quick-fix to update the compile notation 😄)

samtstern:
The final build: "Ran for 6 min 50 sec"

@SUPERCILEX you're a legend, thank you!

SUPERCILEX:
Haha, thanks! 😀

PR #326, opened 2017-09-01
added default channel parameter when constructing own notification

maxost:
Example method `sendNotification` uses deprecated notification builder constructor (without specified notification channel). So if we'll use this method on Android O - notifications won't show up.


maxost:
I've signed CLA

kroikie:
Thanks @maxost

PR #330, opened 2017-09-03
fixed metadata name for default notification channel

maxost:
There is incorrect metadata name for default notification channel. I've posted [stackoverflow question](https://stackoverflow.com/questions/45995337/firebase-messaging-default-notification-channel-doesnt-work) and got answer from Diego Giorgini. So here is fix.


samtstern:
@maxost thank you!  We really appreciate your coming back with a PR.

cc: @kroikie

PR #333, opened 2017-09-05
Update StickerProvider.java

rogerthomas84:
Fixes typo `is not is` to `is not in`


samtstern:
Thanks @rogerthomas84 !

PR #350, opened 2017-10-11
Use sticker builders.

kroikie:
Use sticker builders to define the properties of stickers and
sticker pack.


samtstern:
@kroikie LGTM, squash and merge whenever.

PR #360, opened 2017-10-25
Update to gradle plugin 3.0

samtstern:
Android Studio 3.0 is now stable, therefore we can update this sample.


GitTom:
In the gradle-wrapper.properties I think your intention was just to increment the version, but in some you've made another change, eg.

-distributionUrl=https\://services.gradle.org/distributions/gradle-4.0-all.zip		 +distributionUrl=https\://services.gradle.org/distributions/gradle-4.1-bin.zip

Note that you've changed from 'all' to 'zip'.

PR #397, opened 2017-12-24
Firebase is a mobile platform right? Then why Google didn't made any Firebase Android app?

SFI-tech:
I didn't saw any Firebase suggestions page in their website. That's why I'm posting this at here. Please forward this message to the right place if this is not a right place.
Firebase is a mobile platform right?
Then why Google didn't made any Firebase Android app?
Google is making a lots of apps like Google business, Google docs etc. Don't you thing that Firebase is greater and more useful than Google docs?
There are a lots of guys making their own Firebase Android apps.
But all of 'em has advertisements they are using Firebase for earning money in that way. Why don't you just make an app which is not just a webview but is showing all the Firebase features right in the mobile?


samtstern:
@SFI-tech a pull request is not the right place for this discussion.

If you have a feature request, you can file it here:
https://firebase.google.com/support/contact/bugs-features/

PR #407, opened 2018-01-19
Add dependency version plugin and config

samtstern:


samtstern:
This goes with
https://github.com/GoogleCloudPlatform/repository-gardener/pull/38

cc @kroikie

PR #410, opened 2018-01-21
Fix typo in cities string-array

lineargs:


samtstern:
@lineargs nice catch!

PR #415, opened 2018-01-26
Remove firestore delete operation from client side

### LINKED ISSUE #405: Firestore Sample Friendly Eats: Delete All menu option should actually Delete All
### Step 2: Describe your environment

  * Android device: Pixel running in the emulator on MacOS
  * Android OS version: 5.1.1
  * Firebase/Play Services SDK version: 11.8.0
  
### Step 3: Describe the problem:
Fire Eats has a menu item called "Delete All." When I select this option, the restaurants are deleted from the UI, which is great, but since it only deletes the restaurants and not the ratings (which are sub-collections), if you look at the Firestore db in the console you still see all the implicit restaurant ids and the ratings are still there. I think "Delete All" should actually delete all the data so that it's clean in the app but also clean in the database and the console.

#### Steps to reproduce:
  1. Set up Friendly Eats by following the instructions at https://github.com/firebase/quickstart-android/tree/master/firestore
  2. Start the app. I'm doing it in the emulator but running on a physical device will work fine too.
  3. Under the menu, select "Add Random Items."
  4. Look in the Firestore console. You'll see a collection called restaurants, and under each restaurant a sub-collection called ratings.
  5. Back in the app, under the menu, select "Delete All."
  6. Look in the Firestore console. You'll still see all the restaurant ids under the restaurants collection (grayed out and in italics because the restaurant documents no longer exist), and under each restaurant id you'll still see an actual ratings collection.
  
#### Observed Results:

  * The database is not empty.
  
#### Expected Results:

  * The database should be empty.

samtstern:
This fixes #405

### DIFF
--- a/firestore/README.md
+++ b/firestore/README.md
@@ -23,10 +23,10 @@ Add the following security rules to your project in the:
 service cloud.firestore {  
   match /databases/{database}/documents {
     // Anyone can read a restaurant, only authorized
-    // users can create, update, or delete them.
+    // users can create or update. Deletes are not allowed.
   	 match /restaurants/{restaurantId} {
     	 allow read: if true;
-    	 allow create, update, delete: if request.auth.uid != null;
+    	 allow create, update: if request.auth.uid != null;
     }
     
     // Anyone can read a rating. Only the user who made the rating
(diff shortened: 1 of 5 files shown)


samtstern:
@dpebot merge when green

PR #426, opened 2018-02-09
Since android Oreo notification channel is needed.

Tharunreddych:
Changes to add NotificationChannel as this is needed from Oreo for the notification to appear.


Tharunreddych:
I signed it!

samtstern:
@Tharunreddych thanks for this!  We should have caught it, so glad you did.

PR #454, opened 2018-03-22
Changed android tests to run against release build

ashwinraghav:


functions/app/build.gradle
```diff
@@ -15,9 +18,13 @@ android {
     }
 
     buildTypes {
+
+
         release {
             minifyEnabled true
             proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
+            testProguardFiles getDefaultProguardFile('proguard-android.txt'), 'test-proguard-rules.pro'
```
samtstern: I believe this is missing the final `, 'proguard-rules.pro'` that we discussed

functions/app/build.gradle
```diff
@@ -15,9 +18,13 @@ android {
     }
 
     buildTypes {
+
```
samtstern: nit: random empty lines? (here and above)

functions/app/build.gradle
ashwinraghav: Later figured that was not required. I was running the wrong variant

samtstern:
Hmmm:

```
* What went wrong:
Could not determine the dependencies of task ':functions:app:check'.
> Task with path 'assembleDebugAndroidTest' not found in project ':functions:app'.
```

ashwinraghav:
Seems like there was a dependency on the assembleDebugAndroidTest that crept into the build.
This is thematic across all other modules too.
It is not clear why this dependency was introduced into the 'check' task in any of the projects since I can't find any actual tests.
What was happening in this case is that setting the testBuildType to release resulted in the "assembleDebugAndroidTest" task no longer being defined. See: https://google.github.io/android-gradle-dsl/current/com.android.build.gradle.AppExtension.html#com.android.build.gradle.AppExtension:testBuildType

PR #479, opened 2018-04-25
Changing tests to work on release artifacts.

ashwinraghav:
We also change the number of reviews since there exists a bug on
recycler views where smoothScrollToPosition leaves a gap on top,
leaving the tests unable to see the element which was added.

One solution would be to use well known techniques in espresso [1],
but this requires us to add keep rules for RecyclerView methods
like `scrollToPosition`, since these are used internally by espresso.
If you feel adding the keep rule for RecyclerViews is a reasonable change
to make to the quickstart app, I would prefer that.

This solution on the other hand makes the reviews list small enough where
the gap still leaves the review on screen.

Reference
[1] : https://developer.android.com/training/testing/espresso/lists.html


samtstern:
>  If you feel adding the keep rule for RecyclerViews is a reasonable change
to make to the quickstart app, I would prefer that.

@ashwinraghav that's totally fine with me.

samtstern:
@ashwinraghav also sorry #478 had some overlap here so you'll need to merge.

firestore/app/src/androidTest/java/com/google/firebase/example/fireeats/RecyclerViewMatcher.java
```diff
@@ -0,0 +1,71 @@
...
+
+import android.content.res.Resources;
+import android.support.v7.widget.RecyclerView;
+import android.view.View;
+import org.hamcrest.Description;
+import org.hamcrest.Matcher;
+import org.hamcrest.TypeSafeMatcher;
+
+/**
+ * Created by dannyroa on 5/10/15.
```
samtstern: Probably should ditch this comment.

samtstern:
Approved with one comment.

firestore/app/src/androidTest/java/com/google/firebase/example/fireeats/RecyclerViewMatcher.java
ashwinraghav: Should I ? What's the right way to credit them?

ashwinraghav: The original code is not licensed and is on Github and seems to be used commonly. Left a link like you suggested

samtstern:
@ashwinraghav unfortunately that repo does not have a `LICENSE` file so I don't think we're allowed to copy their code and redistribute under Apache 2.0.   I know that's really annoying, maybe we could contact the author and get the license added?

PR #499, opened 2018-05-14
Add face landmark detection in ML kit demo

thundertrick:
I found ML kit provides a real-time 10-points face landmark detection. Why not add it to the demo?


samtstern:
@thundertrick looks good to me, thank you for contributing!

PR #501, opened 2018-05-16
Convert crash quickstart to crashlytics

### LINKED ISSUE #486: Convert crash reporting sample to crashlytics
Now that Crashlytics is the recommended solution, we should convert the sample here.

samtstern:
Fixes #486

### DIFF
--- a/crash/README.md
+++ b/crash/README.md
@@ -1,10 +1,10 @@
-Firebase Crash Quickstart
-=============================
+Firebase Crashlytics Quickstart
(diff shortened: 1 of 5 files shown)


samtstern:
Note that this will break one snippet include in the docs, I think we can hardcode that one back into those docs.

kroikie:
LGTM, merge once the docs are updated.

samtstern:
cl/196893034

PR #512, opened 2018-05-25
Extend UIAutomator timeouts.

revan:
We're seeing extreme flakiness with 30s timeouts, so this should do the trick. Hopefully.


revan:
On second thought I'll push this through to unblock myself. Will shamefully roll back if needed.

PR #514, opened 2018-05-26
MLKit sample: Remove inappropriate local repository setting

yanzm:
I think this setting is not appropriate because it depends on the directory of a specific user.


samtstern:
@yanzm oops, good catch!

PR #519, opened 2018-05-29
Add waitForIdle call to fix action bar test flakiness.

revan:


revan:
CI failure mentions a timeout to "jcenter.bintray.com:443", so I don't think it's related to this change.

samtstern:
We're seeing this across repos.  Might be a jCenter outage.

PR #526, opened 2018-06-06
Missing link with firebase crash

soulkeykim:
I just try to use sample app, and there is missing link dependencies for firebase crash.
After I added this dependency, I could see crash report from firebase dashboard.

This is error message that I could receive

com.google.firebase.crash.FirebaseCrash is not linked. Skipping initialization.


soulkeykim:
I signed it!

PR #528, opened 2018-06-06
Missing link with firebase crash

soulkeykim:
I just try to use sample app, and there is missing link dependencies for firebase crash. After I added this dependency, I could see crash report from firebase dashboard.

This is error message that I could receive
com.google.firebase.crash.FirebaseCrash is not linked. Skipping initialization.

### DIFF
--- a/crash/app/build.gradle
+++ b/crash/app/build.gradle
@@ -37,6 +37,7 @@ dependencies {
     implementation 'com.android.support:appcompat-v7:27.1.1'
 
     implementation 'com.google.firebase:firebase-core:16.0.0'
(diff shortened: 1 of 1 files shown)


samtstern:
@soulkeykim thanks for sending this Pull Request!  But this sample was recently converted from using Firebase Crash Reporting to Crashlytics, so we intentionally don't have the `firebase-crash` dependency (which is deprecated).

soulkeykim:
Hi @samtstern 

I understand recently changed to only using crashlytics. However, I suggest to set up new project. I was not  able to get the initial dashboard until I add this dependency.

I guess there is an issue that is not expected.

Thank you to look my PR

PR #529, opened 2018-06-07
add firebase core dependency

yanzm:
If this dependency is not added, a warning will be displayed.

screen shot 2018-06-07 at 15 23 50


samtstern:
@yanzm thank you!

PR #530, opened 2018-06-07
Update Dynamic Links sample for custom subdomains

kevinthecheung:


samtstern:
@kevinthecheung thanks for doing this!

PR #556, opened 2018-07-09
Update deps fix

samtstern:
This change updates all dependencies and moves to non-deprecated APIs for Instance ID code.

TODO:

  * Make associated docs changes
  * Wait for `google-services` `4.0.2` to be released

PR #558, opened 2018-07-10
Update FCM app to use onNewToken method added in 17.1.0

morepork:


samtstern:
This is in flight in #556 along with the changes needed to make the build pass.  Waiting on `google-services` version `4.0.2`

samtstern:
But thank you @morepork for the effort!

Issue #557, opened 2018-07-10
[Proposal] Quickstart Kotlin


rosariopfernandes:
It has been more than a year since [Kotlin](https://developer.android.com/kotlin/) became an official language for Android. It is being widely used by Android Developers and most of android samples and blog posts found on the internet are written in Kotlin nowadays.

I suggest rewriting, in Kotlin, the same android samples contained in this repository. Maybe pushing it to a new repository, since the java samples are also helpful.

samtstern:
@rosariopfernandes thanks for bringing this up for discussion!

We are definitely working on bringing Kotlin to our Firebase developer materials, alongside Java.  The first thing we need to do is get our reference docs in order (since those are the most basic thing) and then we can move on to snippets and samples.

the-dagger:
@samtstern Any tips on how folks can help the firebase team with the conversion?
Maybe starting off with the reference docs is something that we can help you out with?

samtstern:
@the-dagger we're actually going to put reference docs on hold for now and jump straight into getting Kotlin snippets alongside the Java snippets in our guides.  I will be working with one of our tech writers int the very near future to start this.

I still have to think of a plan for getting Kotlin into this repo in a maintainable way that still makes it easy to navigate.  In `snippets-android` we defined common interface between Kotlin and Java classes but that doesn't work as well for an app.

the-dagger:
I see, @samtstern that makes sense.
How doing something similar to what the official Android docs do?
They have separate packages for Android and Kotlin for the docs.

<img width="360" alt="screenshot 2018-07-24 at 10 06 03 pm" src="https://user-images.githubusercontent.com/6565270/43153098-fcf7d86c-8f8d-11e8-82d4-87111d98b5c3.png">

This will however be an issue with Dart arrives later on down the line, but happy to brainstorm more on this if you'd like to 👍

samtstern:
@the-dagger yep that's the most likely course.  Can you link me to that repo?  Also what editor/theme are you using?  Very pretty haha.

the-dagger:
Sure, here you go! 
https://github.com/android/snippets

Let me know if there's a way I can help, happy to do so 🙂 

For the theme, I'm using a chrome extension named Octotree, makes it super easy to browser github repos!
https://chrome.google.com/webstore/detail/octotree/bkhaagjahfmjljalopjnoealnfndnagc
[...]

PR #620, opened 2018-09-06
First Kotlin Quickstart

### LINKED ISSUE #633: [Internal] Translate inappmessaging app to kotlin
See #557

samtstern:
See #557

Fixes #633 

@the-dagger
@rosariopfernandes

Here's how it looks:
chooser


inappmessaging/app/src/main/java/com/google/firebase/fiamquickstart/EntryChoiceActivity.java
```diff
@@ -0,0 +1,27 @@
...
+
+import android.content.Intent;
+
+import com.firebase.example.internal.BaseEntryChoiceActivity;
+import com.firebase.example.internal.Choice;
+
+import java.util.Arrays;
+import java.util.List;
+
+public class EntryChoiceActivity extends BaseEntryChoiceActivity {
```
the-dagger: unrelated : can we have this activity in Kotlin for the next quickstarters?

inappmessaging/app/src/main/java/com/google/firebase/fiamquickstart/KotlinMainActivity.kt
```diff
@@ -0,0 +1,55 @@
...
+        super.onCreate(savedInstanceState)
+        setContentView(R.layout.activity_main)
+
+        firebaseAnalytics = FirebaseAnalytics.getInstance(this)
+        firebaseIam = FirebaseInAppMessaging.getInstance()
+
+        firebaseIam.isAutomaticDataCollectionEnabled = true
+        firebaseIam.setMessagesSuppressed(false)
+
+        findViewById<View>(R.id.event_trigger_button)
```
the-dagger: Using synthetic binding here might be a good idea, what do you think?

samtstern: Just learned about a new thing!  Thanks for showing me that.

rosariopfernandes:
Looks better than what I imagined! :clap:

samtstern:
@rosariopfernandes yes we definitely can.  Let's do the simplest ones first and work our way up.

Also so we don't step on each other's toes let's do the following:

  * I'll file an issue for each quickstart
  * Before starting a quickstart translation, comment in the issue that you're going to handle it
  * PRs should only be one quickstart at a time

Thank you for always offering your help!

samtstern:
@rosariopfernandes 

https://github.com/firebase/quickstart-android/milestone/1

PR #641, opened 2018-09-07
feat : convert the auth quickstart to Kotlin

### LINKED ISSUE #625: [Internal] Translate auth app to kotlin
See #557

the-dagger:
@samtstern have a look at the conversion, I have tested it on my end and the converted app works well.
Fixes #625

### DIFF
--- a/admob/build.gradle
+++ b/admob/build.gradle
@@ -2,9 +2,9 @@
 
 buildscript {
     repositories {
-        jcenter()
         mavenLocal()
         google()
+        jcenter()
     }
     dependencies {
         classpath 'com.android.tools.build:gradle:3.1.4'
@@ -17,9 +17,8 @@ buildscript {
 
 allprojects {
     repositories {
-         //mavenLocal() must be listed at the top to facilitate testing
         mavenLocal()
-        jcenter()
         google()
+        jcenter()
     }
 }
--- a/auth/app/build.gradle
+++ b/auth/app/build.gradle
@@ -1,4 +1,6 @@
 apply plugin: 'com.android.application'
+apply plugin: 'kotlin-android'
+apply plugin: 'kotlin-android-extensions'
 check.dependsOn 'assembleDebugAndroidTest'
 
 android {
@@ -36,8 +38,10 @@ configurations.all {
 }
 
 dependencies {
+    implementation project(':internal')
     implementation 'com.android.support:multidex:1.0.3'
 
+    implementation 'com.android.support.constraint:constraint-layout:1.1.3'
     implementation 'com.android.support:appcompat-v7:27.1.1'
     implementation 'com.android.support:animated-vector-drawable:27.1.1'
     implementation 'com.android.support:cardview-v7:27.1.1'
@@ -66,6 +70,10 @@ dependencies {
     androidTestImplementation 'com.android.support.test.espresso:espresso-core:3.0.2'
     androidTestImplementation 'com.android.support.test:rules:1.0.2'
     androidTestImplementation 'com.android.support.test:runner:1.0.2'
+    compile "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
 }
 
 apply plugin: 'com.google.gms.google-services'
+repositories {
+    mavenCentral()
+}
--- a/auth/app/build.gradle
+++ b/auth/app/build.gradle
@@ -74,6 +74,3 @@ dependencies {
 }
 
 apply plugin: 'com.google.gms.google-services'
-repositories {
-    mavenCentral()
-}
--- a/auth/app/build.gradle
+++ b/auth/app/build.gradle
@@ -70,7 +70,7 @@ dependencies {
     androidTestImplementation 'com.android.support.test.espresso:espresso-core:3.0.2'
     androidTestImplementation 'com.android.support.test:rules:1.0.2'
     androidTestImplementation 'com.android.support.test:runner:1.0.2'
-    compile "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
+    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
 }
 
 apply plugin: 'com.google.gms.google-services'
--- a/auth/app/src/main/AndroidManifest.xml
+++ b/auth/app/src/main/AndroidManifest.xml
@@ -62,23 +62,23 @@
             android:name=".java.FirebaseUIActivity"
             android:label="@string/label_firebase_ui" />
 
-        <activity android:name=".kotlin.KotlinChooserActivity" />
+        <activity android:name=".kotlin.ChooserActivity" />
 
         <activity
-            android:name=".kotlin.KotlinCustomAuthActivity"
+            android:name=".kotlin.CustomAuthActivity"
             android:label="@string/label_custom_auth" />
         <activity
-            android:name=".kotlin.KotlinGoogleSignInActivity"
+            android:name=".kotlin.GoogleSignInActivity"
             android:label="@string/label_google_sign_in" />
         <activity
-            android:name=".kotlin.KotlinFacebookLoginActivity"
+            android:name=".kotlin.FacebookLoginActivity"
             android:label="@string/label_facebook_login" />
         <activity
-            android:name=".kotlin.KotlinEmailPasswordActivity"
+            android:name=".kotlin.EmailPasswordActivity"
             android:label="@string/label_emailpassword"
             android:windowSoftInputMode="stateAlwaysHidden" />
         <activity
-            android:name=".kotlin.KotlinPasswordlessActivity"
+            android:name=".kotlin.PasswordlessActivity"
             android:label="@string/label_passwordless"
             android:launchMode="singleTop">
             <intent-filter>
@@ -93,17 +93,17 @@
             </intent-filter>
         </activity>
         <activity
-            android:name=".kotlin.KotlinPhoneAuthActivity"
+            android:name=".kotlin.PhoneAuthActivity"
             android:label="@string/label_phone_auth"
             android:windowSoftInputMode="adjustPan" />
         <activity
-            android:name=".kotlin.KotlinAnonymousAuthActivity"
+            android:name=".kotlin.AnonymousAuthActivity"
             android:label="@string/label_anonymous_auth" />
         <activity
-            android:name=".kotlin.KotlinTwitterLoginActivity"
+            android:name=".kotlin.TwitterLoginActivity"
             android:label="@string/label_twitter_login" />
         <activity
-            android:name=".kotlin.KotlinFirebaseUIActivity"
+            android:name=".kotlin.FirebaseUIActivity"
             android:label="@string/label_firebase_ui" />
 
         <!-- Facebook Configuration -->
--- a/auth/app/src/main/AndroidManifest.xml
+++ b/auth/app/src/main/AndroidManifest.xml
@@ -84,7 +84,7 @@
             <intent-filter>
                 <action android:name="android.intent.action.VIEW" />
                 <data
-                    android:host="auth.example.com"
+                    android:host="kotlin.auth.example.com"
                     android:pathPrefix="/emailSignInLink"
                     android:scheme="https" />
 
--- a/auth/app/src/main/AndroidManifest.xml
+++ b/auth/app/src/main/AndroidManifest.xml
@@ -12,52 +12,98 @@
         android:label="@string/app_name"
         android:supportsRtl="true"
         android:theme="@style/AppTheme">
-        <activity android:name=".ChooserActivity">
+        <activity android:name=".EntryChoiceActivity">
             <intent-filter>
                 <action android:name="android.intent.action.MAIN" />
 
                 <category android:name="android.intent.category.LAUNCHER" />
             </intent-filter>
         </activity>
+        <activity android:name=".java.ChooserActivity" />
         <activity
-            android:name=".CustomAuthActivity"
+            android:name=".java.CustomAuthActivity"
             android:label="@string/label_custom_auth" />
         <activity
-            android:name=".GoogleSignInActivity"
+            android:name=".java.GoogleSignInActivity"
             android:label="@string/label_google_sign_in" />
         <activity
-            android:name=".FacebookLoginActivity"
+            android:name=".java.FacebookLoginActivity"
             android:label="@string/label_facebook_login" />
         <activity
-            android:name=".EmailPasswordActivity"
+            android:name=".java.EmailPasswordActivity"
             android:label="@string/label_emailpassword"
             android:windowSoftInputMode="stateAlwaysHidden" />
         <activity
-            android:name=".PasswordlessActivity"
+            android:name=".java.PasswordlessActivity"
             android:label="@string/label_passwordless"
             android:launchMode="singleTop">
             <intent-filter>
                 <action android:name="android.intent.action.VIEW" />
                 <data
-                    android:scheme="https"
                     android:host="auth.example.com"
-                    android:pathPrefix="/emailSignInLink" />
+                    android:pathPrefix="/emailSignInLink"
+                    android:scheme="https" />
+
                 <category android:name="android.intent.category.BROWSABLE" />
                 <category android:name="android.intent.category.DEFAULT" />
             </intent-filter>
         </activity>
         <activity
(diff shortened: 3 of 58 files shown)


inappmessaging/app/src/main/java/com/google/firebase/fiamquickstart/kotlin/KotlinMainActivity.kt
```diff
@@ -1,11 +1,12 @@
-package com.google.firebase.fiamquickstart
+package com.google.firebase.fiamquickstart.kotlin
```
the-dagger: @samtstern I've made some changes to the inappmessaging module by splitting the code in 2 packages for `java` and `kotlin` each.

admob/build.gradle
```diff
@@ -2,9 +2,9 @@
 
 buildscript {
     repositories {
-        jcenter()
         mavenLocal()
         google()
+        jcenter()
```
the-dagger: Had to do this to get rid of the following error : 
https://issuetracker.google.com/issues?q=status:open%20Could%20not%20find%20play-services-stats.aar

the-dagger:
@samtstern @rosariopfernandes not sure if you faced this, but seems like studio is having a hard time to detect the contents of the `internal` module.

The module builds successfully regardless.
screenshot 2018-09-07 at 8 43 30 pm

auth/app/build.gradle
```diff
@@ -66,6 +70,10 @@ dependencies {
     androidTestImplementation 'com.android.support.test.espresso:espresso-core:3.0.2'
     androidTestImplementation 'com.android.support.test:rules:1.0.2'
     androidTestImplementation 'com.android.support.test:runner:1.0.2'
+    compile "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
 }
 
 apply plugin: 'com.google.gms.google-services'
+repositories {
```
samtstern: What's up with this `repositories` block at the bottom?

auth/app/src/main/AndroidManifest.xml
```diff
@@ -12,52 +12,98 @@
...
+        <activity
+            android:name=".kotlin.KotlinEmailPasswordActivity"
+            android:label="@string/label_emailpassword"
+            android:windowSoftInputMode="stateAlwaysHidden" />
+        <activity
+            android:name=".kotlin.KotlinPasswordlessActivity"
+            android:label="@string/label_passwordless"
+            android:launchMode="singleTop">
+            <intent-filter>
+                <action android:name="android.intent.action.VIEW" />
```
samtstern: Will having two activities with the same intent filter make the quickstart less useful?  Maybe we should do a small code change to use like "kotlin.auth.example.com" in the kt example?

auth/app/src/main/AndroidManifest.xml
```diff
@@ -12,52 +12,98 @@
...
+            android:name=".java.TwitterLoginActivity"
+            android:label="@string/label_twitter_login" />
+        <activity
+            android:name=".java.FirebaseUIActivity"
+            android:label="@string/label_firebase_ui" />
+
+        <activity android:name=".kotlin.KotlinChooserActivity" />
+
+        <activity
+            android:name=".kotlin.KotlinCustomAuthActivity"
```
samtstern: Since all of these classes are in the `.kotlin` package, could we remove the `Kotlin...` name prefix on the class?  That way the names can be 1:1 but different packages.

(I know I did this wrong in the `inappmessaging` sample, that one was so simple that it didn't reveal all the organizational issues that the larger ones do)

auth/app/src/main/java/com/google/firebase/quickstart/auth/java/AnonymousAuthActivity.java
```diff
@@ -81,26 +78,23 @@ private void signInAnonymously() {
...
-                            Log.w(TAG, "signInAnonymously:failure", task.getException());
-                            Toast.makeText(AnonymousAuthActivity.this, "Authentication failed.",
-                                    Toast.LENGTH_SHORT).show();
-                            updateUI(null);
-                        }
-
-                        // [START_EXCLUDE]
-                        hideProgressDialog();
-                        // [END_EXCLUDE]
+                .addOnCompleteListener(this, task -> {
```
samtstern: Ok you might hate me for saying this now ... but can we avoid doing any changes to the Java samples in these PRs?

In general we actually don't want to move the Java samples to Java 8 since I think people who have not moved to Kotlin are also more likely to be stuck on Java 7.

auth/app/src/main/java/com/google/firebase/quickstart/auth/kotlin/KotlinAnonymousAuthActivity.kt
```diff
@@ -0,0 +1,180 @@
...
+import android.widget.TextView
+import android.widget.Toast
+import com.google.android.gms.tasks.OnCompleteListener
+import com.google.android.gms.tasks.Task
+import com.google.firebase.auth.AuthResult
+import com.google.firebase.auth.EmailAuthProvider
+import com.google.firebase.auth.FirebaseAuth
+import com.google.firebase.auth.FirebaseUser
+import com.google.firebase.quickstart.auth.R
+import com.google.firebase.quickstart.auth.java.BaseActivity
```
samtstern: We should make a `BaseActivity.kt` so that the `kotlin` package has no dependencies on the `java` package.

I'd also like to make a linter for the quickstarts to enforce that in the long term.

auth/app/src/main/java/com/google/firebase/quickstart/auth/kotlin/KotlinAnonymousAuthActivity.kt
```diff
@@ -0,0 +1,180 @@
...
+/**
+ * Activity to demonstrate anonymous login and account linking (with an email/password account).
+ */
+class KotlinAnonymousAuthActivity : BaseActivity(), View.OnClickListener {
+
+    // [START declare_auth]
+    private var mAuth: FirebaseAuth? = null
+    // [END declare_auth]
+
+    private var mEmailField: EditText? = null
```
samtstern: For things initialized in `onCreate` I've been going with `lateinit var` fields of non-nullable types (where possible).  WDYT?

auth/app/src/main/java/com/google/firebase/quickstart/auth/kotlin/KotlinChooserActivity.kt
```diff
@@ -0,0 +1,76 @@
...
+ *     {@link FacebookLoginActivity}
+ *     {@link TwitterLoginActivity}
+ *     {@link EmailPasswordActivity}
+ *     {@link PasswordlessActivity}
+ *     {@link PhoneAuthActivity}
+ *     {@link AnonymousAuthActivity}
+ *     {@link KotlinCustomAuthActivity}
+ */
+class KotlinChooserActivity : AppCompatActivity(), AdapterView.OnItemClickListener {
+    private val CLASSES = arrayOf(GoogleSignInActivity::class.java, FacebookLoginActivity::class.java, TwitterLoginActivity::class.java, EmailPasswordActivity::class.java, PasswordlessActivity::class.java, PhoneAuthActivity::class.java, AnonymousAuthActivity::class.java, FirebaseUIActivity::class.java, CustomAuthActivity::class.java)
```
samtstern: nit: can you run the formatter on this class to get rid of the super-long lines?

samtstern: Seems like we're pointing to the original java classes here, shouldn't we be pointing to `KotlinFooActivity`?

auth/app/src/main/java/com/google/firebase/quickstart/auth/kotlin/KotlinAnonymousAuthActivity.kt
the-dagger: > I've been going with lateinit var fields of non-nullable types (where possible)
Ok, makes sense, that's what I've done in other activities as well.
Seems like, I left the views here by mistake, have been using synthetic binding for them wherever possible.
Will edit this one out!

auth/app/src/main/java/com/google/firebase/quickstart/auth/kotlin/KotlinChooserActivity.kt
the-dagger: Sure

the-dagger: My bad! 😖

auth/app/src/main/java/com/google/firebase/quickstart/auth/kotlin/KotlinAnonymousAuthActivity.kt
samtstern: Yep synthetic binding is awesome!   But also fine with fields for views in places where it's awkward.  Also feel free to refactor the common view IDs to be camelCase if it makes synthetic binding look more natural (I think I used underscores in most of the IDs)

auth/app/src/main/AndroidManifest.xml
the-dagger: > Maybe we should do a small code change to use like "kotlin.auth.example.com" in the kt example?

Sorry, I didn't seem to understand this one!
Could you outline a bit more on what you meant here?
Thanks!

auth/app/src/main/java/com/google/firebase/quickstart/auth/kotlin/KotlinAnonymousAuthActivity.kt
the-dagger: Yes, I'll do that.

auth/app/build.gradle
rosariopfernandes: I had this in my ML Kit translation... It was auto-generated by Android Studio when I used their assistant to add Kotlin to the project. I undone the changes and added Kotlin manually.

Related: What if we declare the kotlin version as a variable on the root build.gradle file and then call it in every module? (like we have on the snippets repo)

auth/app/src/main/AndroidManifest.xml
rosariopfernandes: @the-dagger that intent filter handles calls to "auth.example.com". Two activities can't handle the same URL. One of them will be picked instead of the other. That's why @samtstern recommends using "kotlin.auth.example.com" for the Kotlin Activity.

the-dagger: Ah, gotcha!
Thanks :)

the-dagger:
@samtstern Thanks for the detailed review, done with the changes!

auth/app/src/main/java/com/google/firebase/quickstart/auth/kotlin/AnonymousAuthActivity.kt
```diff
@@ -0,0 +1,163 @@
...
+
+    private fun signInAnonymously() {
+        showProgressDialog()
+        // [START signin_anonymously]
+        mAuth.signInAnonymously()
+                .addOnCompleteListener(this) { task ->
+                    if (task.isSuccessful) {
+                        // Sign in success, update UI with the signed-in user's information
+                        Log.d(TAG, "signInAnonymously:success")
+                        val user = mAuth!!.currentUser
```
samtstern: I think this `!!` is no longer necessary now that you've moved it to a non-nullable `lateinit var`.

I suspect there are probably a lot of these floating around, so just gonna comment on it once but please fix as many as you can.

In my Android Studio it highlights them in yellow, but I can't find a "fix all" option.

auth/app/src/main/java/com/google/firebase/quickstart/auth/kotlin/AnonymousAuthActivity.kt
```diff
@@ -0,0 +1,163 @@
...
+        if (!validateLinkForm()) {
+            return
+        }
+
+        // Get email and password from the form
+        val email = fieldEmail.text.toString()
+        val password = fieldPassword.text.toString()
+
+        // Create EmailAuthCredential with email and password
+        val credential = EmailAuthProvider.getCredential(email.toString(), password.toString())
```
samtstern: nit: Do we need to call `toString()` on these?  I think we did that in the `val` declaration.

auth/app/src/main/java/com/google/firebase/quickstart/auth/kotlin/AnonymousAuthActivity.kt
```diff
@@ -0,0 +1,163 @@
...
+
+        // Link the anonymous user to the email credential
+        showProgressDialog()
+
+        // [START link_credential]
+        mAuth.currentUser?.linkWithCredential(credential)
+                ?.addOnCompleteListener(this) { task ->
```
samtstern: Does `linkWithCredential` need to be annotated to say it returns a `@NonNull` `Task` object?  If so I can file a bug against the auth team.

In general if you find ways that our Java APIs are less-than-perfect in Kotlin, keep a list and send it to me and I'll poke people to fix it.

auth/app/src/main/java/com/google/firebase/quickstart/auth/kotlin/AnonymousAuthActivity.kt
```diff
@@ -0,0 +1,163 @@
...
+        return valid
+    }
+
+    private fun updateUI(user: FirebaseUser?) {
+        hideProgressDialog()
+        val isSignedIn = user != null
+
+        // Status text
+        if (isSignedIn) {
+            anonymousStatusId.text = getString(R.string.id_fmt, user!!.uid)
```
samtstern: Huh it's lame that the kotlin compiler can't tell that `user != null` in this block because we already check `isSignedIn`.

(just a note, no action needed for you)

auth/app/src/main/java/com/google/firebase/quickstart/auth/kotlin/FacebookLoginActivity.kt
```diff
@@ -0,0 +1,150 @@
...
+        updateUI(null)
+    }
+
+    private fun updateUI(user: FirebaseUser?) {
+        hideProgressDialog()
+        if (user != null) {
+            status.text = getString(R.string.facebook_status_fmt, user.displayName)
+            detail.text = getString(R.string.firebase_status_fmt, user.uid)
+
+            findViewById<View>(R.id.buttonFacebookLogin).visibility = View.GONE
```
samtstern: nit: more synthetic binding usage here?

auth/app/src/main/java/com/google/firebase/quickstart/auth/kotlin/PasswordlessActivity.kt
```diff
@@ -0,0 +1,214 @@
...
+    /**
+     * Send an email sign-in link to the specified email.
+     */
+    private fun sendSignInLink(email: String) {
+        val settings = ActionCodeSettings.newBuilder()
+                .setAndroidPackageName(
+                        packageName,
+                        false, null/* minimum app version */)/* install if not available? */
+                .setHandleCodeInApp(true)
+                .setUrl("https://auth.example.com/emailSignInLink")
```
samtstern: I think you need to add the "kotlin." here to match what you did in the AndroidManifest

auth/app/src/main/java/com/google/firebase/quickstart/auth/kotlin/AnonymousAuthActivity.kt
```diff
@@ -0,0 +1,163 @@
...
+import com.google.firebase.auth.FirebaseAuth
+import com.google.firebase.auth.FirebaseUser
+import com.google.firebase.quickstart.auth.R
+import kotlinx.android.synthetic.main.activity_anonymous_auth.*
+
+/**
+ * Activity to demonstrate anonymous login and account linking (with an email/password account).
+ */
+class AnonymousAuthActivity : BaseActivity(), View.OnClickListener {
+    private val TAG = "AnonymousAuth"
```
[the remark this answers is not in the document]
the-dagger: I have a neutral opinion on this, but making the constants a companion object doesn't seem like a bad idea.

the-dagger:
@samtstern done!

samtstern:
@the-dagger this is great!  Have you done a sanity check to make sure the whole things runs as expected?  I'll trust your assessment there.  If you haven't, happy to do it for you.

samtstern:
:ship:

PR #644, opened 2018-09-07
Converts storage quickstart to Kotlin

### LINKED ISSUE #637: [Internal] Translate storage app to kotlin
See #557

samtstern:
Fixes #637

### DIFF
--- a/storage/app/src/main/java/com/google/firebase/quickstart/firebasestorage/kotlin/MainActivity.kt
+++ b/storage/app/src/main/java/com/google/firebase/quickstart/firebasestorage/kotlin/MainActivity.kt
@@ -79,9 +79,9 @@ class MainActivity : AppCompatActivity(), View.OnClickListener {
         }
 
         // Restore instance state
-        if (savedInstanceState != null) {
-            fileUri1 = savedInstanceState.getParcelable(KEY_FILE_URI)
-            downloadUrl = savedInstanceState.getParcelable(KEY_DOWNLOAD_URL)
+        savedInstanceState?.let {
+            fileUri1 = it.getParcelable(KEY_FILE_URI)
+            downloadUrl = it.getParcelable(KEY_DOWNLOAD_URL)
         }
         onNewIntent(intent)
     }
--- a/storage/app/src/main/java/com/google/firebase/quickstart/firebasestorage/kotlin/MainActivity.kt
+++ b/storage/app/src/main/java/com/google/firebase/quickstart/firebasestorage/kotlin/MainActivity.kt
@@ -1,0 +1,285 @@
+package com.google.firebase.quickstart.firebasestorage.kotlin
+
+import android.app.Activity
+import android.app.ProgressDialog
+import android.content.BroadcastReceiver
+import android.content.Context
+import android.content.Intent
+import android.net.Uri
+import android.os.Bundle
+import android.support.v4.content.LocalBroadcastManager
+import android.support.v7.app.AlertDialog
+import android.support.v7.app.AppCompatActivity
+import android.util.Log
+import android.view.Menu
+import android.view.MenuItem
+import android.view.View
+import android.widget.TextView
+import android.widget.Toast
+import com.google.firebase.auth.FirebaseAuth
+import com.google.firebase.auth.FirebaseUser
+import com.google.firebase.quickstart.firebasestorage.R
+import com.google.firebase.quickstart.firebasestorage.java.MyDownloadService
+import com.google.firebase.quickstart.firebasestorage.java.MyUploadService
+import java.util.*
+
+/**
+ * Activity to upload and download photos from Firebase Storage.
+ *
+ * See [MyUploadService] for upload example.
+ * See [MyDownloadService] for download example.
+ */
+class MainActivity : AppCompatActivity(), View.OnClickListener {
+
+    private var mBroadcastReceiver: BroadcastReceiver? = null
+    private var mProgressDialog: ProgressDialog? = null
+    private var mAuth: FirebaseAuth? = null
+
+    private var mDownloadUrl: Uri? = null
+    private var mFileUri: Uri? = null
+
+    override fun onCreate(savedInstanceState: Bundle?) {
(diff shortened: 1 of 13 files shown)


storage/app/src/main/java/com/google/firebase/quickstart/firebasestorage/kotlin/MainActivity.kt
```diff
@@ -0,0 +1,279 @@
...
+                    MyUploadService.UPLOAD_COMPLETED, MyUploadService.UPLOAD_ERROR -> onUploadResultIntent(intent)
+                }
+            }
+        }
+
+        // Restore instance state
+        if (savedInstanceState != null) {
+            fileUri1 = savedInstanceState.getParcelable(KEY_FILE_URI)
+            downloadUrl = savedInstanceState.getParcelable(KEY_DOWNLOAD_URL)
+        }
```
rosariopfernandes: Maybe use:
```kotlin
savedInstanceState?.let {
    fileUri1 = it.getParcelable(KEY_FILE_URI)
    downloadUrl = it.getParcelable(KEY_DOWNLOAD_URL)
}
```

storage/app/src/main/java/com/google/firebase/quickstart/firebasestorage/kotlin/MainActivity.kt
```diff
@@ -0,0 +1,279 @@
...
+    }
+
+    override fun onActivityResult(requestCode: Int, resultCode: Int, data: Intent) {
+        Log.d(TAG, "onActivityResult:$requestCode:$resultCode:$data")
+        if (requestCode == RC_TAKE_PICTURE) {
+            if (resultCode == Activity.RESULT_OK) {
+                fileUri1 = data.data
+
+                if (fileUri1 != null) {
+                    uploadFromUri(fileUri1!!)
```
the-dagger: Can use the let block here to replace the null check

storage/app/src/main/java/com/google/firebase/quickstart/firebasestorage/kotlin/MainActivity.kt
samtstern: Slick!  I like it

storage/app/src/main/java/com/google/firebase/quickstart/firebasestorage/kotlin/MainActivity.kt
```diff
@@ -0,0 +1,279 @@
...
+                .putExtra(MyUploadService.EXTRA_FILE_URI, fileUri)
+                .setAction(MyUploadService.ACTION_UPLOAD))
+
+        // Show loading spinner
+        showProgressDialog(getString(R.string.progress_uploading))
+    }
+
+    private fun beginDownload() {
+        // Get path
+        val path = "photos/" + fileUri1!!.lastPathSegment
```
the-dagger: Either we can use ?. operator or use let block here to ensure that path is initialised only when the fileUri1 isn't null.

storage/app/src/main/java/com/google/firebase/quickstart/firebasestorage/kotlin/MyBaseTaskService.kt
```diff
@@ -0,0 +1,118 @@
...
+        createDefaultChannel()
+        val builder = NotificationCompat.Builder(this, CHANNEL_ID_DEFAULT)
+                .setSmallIcon(R.drawable.ic_file_upload_white_24dp)
+                .setContentTitle(getString(R.string.app_name))
+                .setContentText(caption)
+                .setProgress(100, percentComplete, false)
+                .setOngoing(true)
+                .setAutoCancel(false)
+
+        val manager = getSystemService(Context.NOTIFICATION_SERVICE) as NotificationManager
```
the-dagger: Maybe make this an instance variable and initialise it ```by lazy {}```

storage/app/src/main/java/com/google/firebase/quickstart/firebasestorage/kotlin/MainActivity.kt
```diff
@@ -0,0 +1,279 @@
...
+            layoutSignin.visibility = View.GONE
+            layoutStorage.visibility = View.VISIBLE
+        } else {
+            layoutSignin.visibility = View.VISIBLE
+            layoutStorage.visibility = View.GONE
+        }
+
+        // Download URL and Download button
+        if (downloadUrl != null) {
+            pictureDownloadUri.text = downloadUrl!!.toString()
```
the-dagger: Not sure if toString() will be needed with !!, also probably kotlin with do a smart inference here since we did the null check already.

storage/app/src/main/java/com/google/firebase/quickstart/firebasestorage/kotlin/MainActivity.kt
rosariopfernandes: @the-dagger I'm not sure about that, since `Log.w()` must be called if the `fileUri1` is indeed null.

the-dagger: Hm, yeah that makes sense.
Anyways, since we are checking for the nullability of fileUri1, it might be a good idea to remove !! from the next line (considering that the compiler is ok with it)

samtstern: It's actually not OK with it because `fileUri` is a nullable instance variable and it "could be changed by this time".  I guess it's worried about concurrency.

the-dagger: Hm, I see

storage/app/src/main/java/com/google/firebase/quickstart/firebasestorage/kotlin/MainActivity.kt
samtstern: Neither of those works but I added an if(){return}

PR #647, opened 2018-09-08
feat : kotlinize crash quickstart

### LINKED ISSUE #627: [Internal] Translate crash app to kotlin
See #557

the-dagger:
fix #627

### DIFF
--- a/crash/app/src/main/java/com/google/samples/quickstart/crash/kotlin/MainActivity.kt
+++ b/crash/app/src/main/java/com/google/samples/quickstart/crash/kotlin/MainActivity.kt
@@ -1,0 +1,75 @@
+package com.google.samples.quickstart.crash.kotlin
+
+import android.os.Bundle
+import android.support.v7.app.AppCompatActivity
+import android.util.Log
+import com.crashlytics.android.Crashlytics
+import com.google.samples.quickstart.crash.R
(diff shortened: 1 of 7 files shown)


crash/app/src/main/java/com/google/samples/quickstart/crash/kotlin/MainActivity.kt
```diff
@@ -0,0 +1,75 @@
...
+        // Add some custom values and identifiers to be included in crash reports
+        Crashlytics.setInt("MeaningOfLife", 42)
+        Crashlytics.setString("LastUIAction", "Test value")
+        Crashlytics.setUserIdentifier("123456789")
+
+        // Report a non-fatal exception, for demonstration purposes
+        Crashlytics.logException(Exception("Non-fatal exception: something went wrong!"))
+
+        // Checkbox to indicate when to catch the thrown exception.
+        val catchCrashCheckBox = catchCrashCheckBox
```
samtstern: Any reason for this `val` to exist?

the-dagger: Yes, for sanity.
Could have skipped it altogether and used synthetic binding but thought that it might be confusing to the user so left it there.

LMK, it that makes sense otherwise I'll remove it.

PR #651, opened 2018-09-10
Perfmon kotlin

### LINKED ISSUE #636: [Internal] Translate perf app to kotlin
See #557

rosariopfernandes:
Fixes #636

### DIFF
--- a/perf/app/src/main/java/com/google/firebase/quickstart/perfmon/kotlin/MainActivity.kt
+++ b/perf/app/src/main/java/com/google/firebase/quickstart/perfmon/kotlin/MainActivity.kt
@@ -1,0 +1,180 @@
+package com.google.firebase.quickstart.perfmon.kotlin
+
+import android.graphics.drawable.ColorDrawable
+import android.os.AsyncTask
+import android.os.Bundle
+import android.support.v4.content.ContextCompat
+import android.support.v7.app.AppCompatActivity
+import android.util.Log
+import android.widget.Button
+import android.widget.Toast
+import com.bumptech.glide.Glide
+import com.bumptech.glide.load.resource.drawable.GlideDrawable
+import com.bumptech.glide.request.RequestListener
+import com.bumptech.glide.request.target.Target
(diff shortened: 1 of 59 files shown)


samtstern:
:ship: :shipit:

perf/app/src/main/java/com/google/firebase/quickstart/perfmon/kotlin/MainActivity.kt
```diff
@@ -0,0 +1,180 @@
...
+
+    private fun loadStringFromFile(): Task<String> {
+        return Tasks.call(
+                AsyncTask.THREAD_POOL_EXECUTOR,
+                Callable {
+                    val contentFile = File(filesDir, CONTENT_FILE)
+                    if (contentFile.createNewFile()) {
+                        // Content file exist did not exist in internal storage and new file was created.
+                        // Copy in the default content.
+                        val `is`: InputStream = assets.open(DEFAULT_CONTENT_FILE)
```
rosariopfernandes: @samtstern the `is` variable here is between grave accent marks because `is` is a reserved word in Kotlin.
I was thinking about changing the variable name for something like `inputStream` to avoid using the grave accent marks.

But I wasn't sure whether we could do that or not. Let me know what you think.

samtstern: I saw that too.  I agree, changing the variable name is probably cleaner.  And feel free to do that in general!  For instance I have been removing the `m` prefix from variables in my kotlin translations.

rosariopfernandes: Does that imply changing the variable name in java as well?

samtstern: No I don't think so, it's fine for Kotlin and Java to have different styles.  In Java it's very common to call an `InputStream` variable `is` so we should leave it in Java.

PR #658, opened 2018-09-11
Implement lint check for java imports

### LINKED ISSUE #643: [Internal] Create a lint to make sure java classes are not imported from kotlin classes
When converting to kotlin it's very easy to get crossed imports.

If we see a package `foo.kotlin`, none of the code in there should import from `foo.java`

the-dagger:
@samtstern I've implemented the lint check successfully as you outlined in the issue.

screenshot 2018-09-11 at 7 21 44 pm

However, currently it checks all the Java and Kotlin files whereas I want to limit this to Kotlin files only.

Any ideas as to how this can be done?
Also would appreciate comments regarding the project structure

Fix #643

### DIFF
--- a/internal/lint/src/test/java/com/firebase/lint/InvalidImportDetectorTest.kt
+++ b/internal/lint/src/test/java/com/firebase/lint/InvalidImportDetectorTest.kt
@@ -7,23 +7,23 @@ import org.junit.Test
 class InvalidImportDetectorTest {
 
     private val javaPackage = java("""
-      |package com.google.firebase.java;
-      |
-      |public final class Hello {
-      |  public static final class drawable {
-      |  }
-      |}""".trimMargin())
+      package com.google.firebase.java;
+
+      public final class Hello {
+        public static final class drawable {
+        }
+      }""").indented()
 
     @Test
     fun normalRImport() {
         lint()
                 .files(javaPackage, java("""
-          |package com.google.firebase.kotlin;
-          |
-          |import com.google.firebase.Hello;
-          |
-          |class Example {
-          |}""".trimMargin()))
+          package com.google.firebase.kotlin;
+
+          import com.google.firebase.Hello;
+
+          class Example {
+          }""").indented())
                 .issues(ISSUE_INVALID_IMPORT)
                 .run()
                 .expectClean()
@@ -33,12 +33,12 @@ class InvalidImportDetectorTest {
     fun wrongImport() {
         lint()
                 .files(javaPackage, java("""
-          |package com.google.firebase.kotlin;
-          |
-          |import com.google.firebase.java.Hello;
-          |
-          |class Example {
-          |}""".trimMargin()))
+          package com.google.firebase.kotlin;
+
+          import com.google.firebase.java.Hello;
+
+          class Example {
+          }""").indented())
                 .issues(ISSUE_INVALID_IMPORT)
                 .run()
                 .expect("""


internal/chooser/src/main/java/com/firebase/example/internal/ChoiceAdapter.java
```diff
@@ -0,0 +1,68 @@
+package com.firebase.example.internal;
```
samtstern: Strange that git can't tell the `chooser` files are just renames and not adds :man_shrugging:

internal/lint/src/main/java/com/firebase/lint/InvalidImportDetector.kt
```diff
@@ -0,0 +1,40 @@
...
+
+    override fun createUastHandler(context: JavaContext) = InvalidImportHandler(context)
+
+    class InvalidImportHandler(private val context: JavaContext) : UElementHandler() {
+
+        override fun visitImportStatement(node: UImportStatement) {
+
+            node.importReference?.let { importReference ->
+
+                if (importReference.asSourceString().contains(disallowedImports)) {
```
samtstern: I wonder if we could be more exact here (excuse my pseudocode):

```kt
// Something like "com.google.firebase.example.java.MyClass"
var importedPackage = importReference.asSourceString();

// The class being analyzed, something like "com.google.firebase.example.kotlin.MyOtherClass"
var callingPackage = ...;

// Now find the first difference between the two packages
var importedSegments = importedPackage.split(".");
var callingSegments = callingPackage.split(".");

var i = 0;
while (i < importedSegments.length && i < callingSegments.length) {
    if (importedSegments[i] == "java"  && callingSegments[i] == "kotlin") {
      // LINT ERROR!
    }
}
```

The logic above only makes it an error for a kotlin class to import a java class from the same quickstart, rather than to import any class with `.java.` in the full path.  Wdyt?

internal/chooser/src/main/java/com/firebase/example/internal/ChoiceAdapter.java
the-dagger: Yeah, that's what I thought too!
Still I built and ran the app to make sure that it works well, and it does.

internal/lint/src/main/java/com/firebase/lint/InvalidImportDetector.kt
the-dagger: Done!

internal/lintchecks/build.gradle
```diff
@@ -0,0 +1,28 @@
...
+
+android {
+    compileSdkVersion 28
+    defaultConfig {
+        minSdkVersion 16
+        targetSdkVersion 28
+        testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
+    }
+
+    compileOptions {
```
vanniktech: also here we don't need java 8

internal/lintchecks/build.gradle
```diff
@@ -0,0 +1,28 @@
+apply plugin: 'com.android.library'
+
+android {
+    compileSdkVersion 28
+    defaultConfig {
+        minSdkVersion 16
+        targetSdkVersion 28
+        testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
```
vanniktech: can be removed too

internal/lintchecks/build.gradle
```diff
@@ -0,0 +1,28 @@
+apply plugin: 'com.android.library'
+
+android {
+    compileSdkVersion 28
+    defaultConfig {
+        minSdkVersion 16
+        targetSdkVersion 28
```
vanniktech: not needed

settings.gradle
```diff
@@ -1 +1,4 @@
-include ':admob:app', ':analytics:app', ':app-indexing:app', ':auth:app', ':config:app', ':crash:app', ':database:app', ':dynamiclinks:app', ':firestore:app', ':functions:app', ':invites:app', ':internal', 'inappmessaging:app', ':messaging:app', ':mlkit:app', ':perf:app', ':storage:app'
+include ':admob:app', ':analytics:app', ':app-indexing:app', ':auth:app', ':config:app',
```
vanniktech: one per line?

the-dagger: Yeah, changed it because it became difficult to manually manage added/removed modules

internal/lint/src/main/java/com/firebase/lint/InvalidImportDetector.kt
```diff
@@ -0,0 +1,53 @@
...
+
+    override fun getApplicableUastTypes() = listOf(UImportStatement::class.java)
+
+    override fun createUastHandler(context: JavaContext) = InvalidImportHandler(context)
+
+    class InvalidImportHandler(private val context: JavaContext) : UElementHandler() {
+
+        override fun visitImportStatement(node: UImportStatement) {
+            var importedPackageName = ""
+            val classPackageName = (context.psiFile as PsiJavaFile).packageName
```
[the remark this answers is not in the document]
the-dagger: What would you recommend?
I want to get the package name of the file and `psiFile` doesn't have the packageName getter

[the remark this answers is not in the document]
vanniktech: `((context.psiFile as PsiFile).firstChild as PsiPackageStatement).packageName`

although firstChild probably not the best since in Kotlin the first statement can be `@file:Suppress("Something")`. So probably go step by step over the children and then look for the first `PsiPackageStatement`

internal/lint/src/main/java/com/firebase/lint/InvalidImportDetector.kt
```diff
@@ -0,0 +1,53 @@
...
+
+            val classPackageSubFolders = classPackageName.split(".")
+            val importedPackageSubFolders = importedPackageName.split(".")
+
+            var i = 0
+            while (i < classPackageSubFolders.size && i < importedPackageSubFolders.size) {
+
+                if (classPackageSubFolders[i] == "java" && importedPackageSubFolders[i] == "kotlin") {
+                    node.importReference?.let {
+                        context.report(ISSUE_INVALID_IMPORT, node, context.getLocation(it), "Invalid import")
```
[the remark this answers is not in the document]
samtstern: In this case we just want the author to remove it.

internal/lint/src/main/java/com/firebase/lint/InvalidImportDetector.kt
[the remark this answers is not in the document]
samtstern: @the-dagger yeah for now we can make that assumption.  You could even enforce that via lint!

the-dagger:
@samtstern @vanniktech I think this should be good to merge now!

internal/lint/src/main/java/com/firebase/lint/InvalidImportDetector.kt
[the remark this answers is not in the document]
the-dagger: Haha, too much lint for the day! 😆 
I'll probably pick this up later.

internal/chooser/src/main/java/com/firebase/example/internal/BaseEntryChoiceActivity.java
```diff
@@ -0,0 +1,25 @@
+package com.firebase.example.internal;
```
samtstern: I just realized: if GitHub thinks this is an 'add' shouldn't there also be a matching delete for the file in the other location?

I worry that after merging this we will have duplicated code.

samtstern:
@vanniktech thanks so much for your review here!

internal/lintchecks/build.gradle
```diff
@@ -2,27 +2,8 @@ apply plugin: 'com.android.library'
 
 android {
     compileSdkVersion 28
-    defaultConfig {
-        minSdkVersion 16
```
vanniktech: This you'll need

internal/lint/src/test/java/com/firebase/lint/InvalidImportDetectorTest.kt
```diff
@@ -0,0 +1,50 @@
+package com.firebase.lint
+
+import com.android.tools.lint.checks.infrastructure.LintDetectorTest.java
+import com.android.tools.lint.checks.infrastructure.TestLintTask.lint
+import org.junit.Test
+
+class InvalidImportDetectorTest {
+
+    private val javaPackage = java("""
+      |package com.google.firebase.java;
```
vanniktech: Remove the `|` and the trimMargin call. On the return type from Java you can call intended() 

The advantage is that you'll get syntax highlighting.

internal/chooser/src/main/java/com/firebase/example/internal/BaseEntryChoiceActivity.java
the-dagger: @samtstern I think I got the issue here.
Seems like my git was acting up, after a system reboot, it started showing the deleted files.

I've committed them, should be all good now!

vanniktech:
@samstern anything when it comes to Lint checks. Just add me as a reviewer. Api isn't great and there are many pitfalls.

the-dagger:
@samtstern the build here is failing because the analytics module merged earlier has a dependency on `:internal` whereas this PR refactors it to `:internal:chooser`.

I can send a PR refactoring this in the conflicting modules after this PR is merged.

samtstern:
Thanks @vanniktech for pointing that out.  @the-dagger I will fix the build.

PR #655, opened 2018-09-10
feat: Quickstart Analytics in Kotlin

rosariopfernandes:
Fixes #623

PR #660, opened 2018-09-11
style: minor style changes in MainActivity.kt

rosariopfernandes:
Implements the changes proposed in the #655 review.

### DIFF
--- a/analytics/app/src/main/java/com/google/firebase/quickstart/analytics/kotlin/MainActivity.kt
+++ b/analytics/app/src/main/java/com/google/firebase/quickstart/analytics/kotlin/MainActivity.kt
@@ -1,0 +1,251 @@
+package com.google.firebase.quickstart.analytics.kotlin
+
+import android.content.Intent
+import android.os.Bundle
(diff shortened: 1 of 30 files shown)


analytics/app/src/main/java/com/google/firebase/quickstart/analytics/kotlin/MainActivity.kt
```diff
@@ -51,7 +50,7 @@ class MainActivity : AppCompatActivity() {
     /**
      * The user's favorite food, chosen from a dialog.
      */
-    private var mFavoriteFood: String? = null
+    private lateinit var mFavoriteFood: String
```
samtstern: Looking at it now ... is `mFavoriteFood` ever used outside of one function?  Could be just a local var in `setUserFavoriteFood`.

rosariopfernandes: You're right. I'll update

PR #661, opened 2018-09-11
Update Android Gradle Plugin to 3.3.0-alpha08. This causes the google_app_id not to be generated.

vanniktech:


rosariopfernandes:
The Android Gradle Plugin version 3.1.4 works fine on my side. Which version of Android Studio are you using? @vanniktech

samtstern:
@rosariopfernandes I think this is a specific issue with `3.3.0-alpha08` combined with `google-services` version `4.1.0`.  I am gonna check it out on a clean project.

vanniktech:
Haven't tried out alpha09 which just got released today. But yes AGP 3.3.0-alpha08 and Google Play Services Plugin 4.1.0 do not play along nicely. Namely, it does not generate the xml resource file containing the extracted information from the `play-services.json` file.

The temporary workaround is to use 4.0.1 or 4.0.2 which does work with the 3.3.0-alpha versions.

samtstern:
@vanniktech was able to reproduce!  Doesn't look like the plugin even runs, since I can apply the `4.1.0` plugin to a project with no `json` file and have the build succeed.

Thanks for this!  I will flag it to the right people.

samtstern:
Plugin owner says: "Yes, I had someone report it earlier this week, I've been investigating"

@vanniktech if you're still in contact with Firebase support you can tell them this is a duplicate of bug 114445730

the-dagger:
Unrelated : @vanniktech Thanks a ton for your [samples on lint](https://github.com/vanniktech/lint-rules), they helped a lot while I was working on #658! 🎉

vanniktech:
Sent it back to Firebase support. Thanks for helping out.

vanniktech:
@the-dagger I've reviewed your PR here about lint.

the-dagger:
Thanks @vanniktech.
Added some followup comments, have a look whenever convenient!

samtstern:
Going to close this as we have an internal repro, so leaving this PR open doesn't do that much.  Will update when we know more!

b95505017:
Any updates? 3.3 is going to beta now, and still can't work with 4.1.0.

LouisCAD:
@samtstern Is there an issue we can follow?

eygraber:
Google services plugin 4.2.0 fixes the problem for me

samtstern:
@eygraber nice!  Thanks for following up.

@LouisCAD no there is not, but if you get the same issues with `4.2.0` please file a new issue!

PR #662, opened 2018-09-11
Fix lint build

samtstern:
Fixes build after #658


samtstern:
@dpebot merge when green

PR #693, opened 2018-10-04
Slight optimize in  BarcodeScanningProcessor.kt

XinyueZ:


samtstern:
The code LGTM.  @rosariopfernandes WDYT?

rosariopfernandes:
@samtstern :ship: ship it!

Oh, and btw: I remember bumping into similar classes (where the constructor is only calling `getInstance()` from the Firebase SDKs). Should we change those as well so that it matches this one?

samtstern:
@rosariopfernandes nah `getInstance()` is not expensive, no reason to rush to optimize if we don't have other changes to make.

PR #696, opened 2018-10-06
Fix lint warnings and optimize for CloudImageLabelingProcessor

XinyueZ:
- Lint-warning at `onSuccess()`
- `detector` with `by-lazy`. 
  - I know people will mind there's no need for performance improvement here, however, for enthusiast :)


samtstern:
@XinyueZ no objections from me, thanks!

PR #727, opened 2018-11-04
[OPTIMIZE]Slight optimize the BarcodeGraphic.kt

XinyueZ:


samtstern:
@XinyueZ thanks for all these pull requests!  In the future it's probably better to merge them into a few larger ones, it will save you a lot of pain with possible reviews and merge conflicts.  Plus it helps me see the bigger picture.

PR #746, opened 2018-11-18
Nullability issues running Kotlin version of Cloud Document Text Recognition

joaobiriba:
Running the Kotlin version of Cloud Document Text Recognition loading an image from the file system I spotted this issue:
`detectInVisionImage` of `VisionProcessorBase` is always called with null `originalCameraImage` and null `metadata`
The metadata check for nullability is always failing and is not executing the `onSuccess` even though we don't really need it in the onSuccess body.

The application is ok using the fix proposed, but proably we needed a better refactor to solve it.


samtstern:
Assigning to @gkaldev ... the code looks fine in isolation but not familiar with the larger issue it's addressing.

joaobiriba:
@gkaldev Any news about this?

gkaldev:
I apologize for the massive delay on this--I somehow just muted this thread. Not sure if something changed in the SDK since this PR was initially created, but I tested just now and it seems to work as it is. I successfully used a file system image to run cloud doc text rec.

Issue #748, opened 2018-11-22
[mlkit] Single Camera Handling


VarPete:
On devices which only have a single (typically front-facing) camera (Pixelbook, Android Things Dev Kit, etc.), the Live Preview attempts to use the back-facing camera on launch but fails to locate one.  

When the user then toggles the camera to move to the front facing one, the user needs to pause the application and return to re-initialize the `CameraSource` and get the preview functional.  

When the preview is functional, clicking the toggle button a second time causes the preview to freeze, and the system to attempt to load/use the non-existent back facing camera.

In single camera configurations, it would be possible to 'short circuit' the `createCamera` or the `getIdForRequestedCamera`, but those optimizations would ignore the `facing` variable, which is used to calculate where to place overlays (namely Facial Detection/Contour) since images are 'mirrored' when camera is front-facing.

The solution proposed in f307e51204256388c7f7f4ec751ce70ba402e404: 

*  Detects that there is a single camera on the device and removes the toggle button from the UI when creating the `LivePreview` activity (both Java and Kotlin)
*  Detects single camera configuration within `CameraSource` and forces the `facing` variable to be equal to the value within `CameraInfo`

samtstern:
Fixed by #749

PR #749, opened 2018-11-22
[mlkit] Single Camera Handling

VarPete:
Proposed solution for #748 

*  Detects that there is a single camera on the device and removes the toggle button from the UI when creating the LivePreview activity (both Java and Kotlin)
*  Detects single camera configuration within CameraSource and forces the facing variable to be equal to the value within CameraInfo

Camera enumeration and meta-data require `android.permission.CAMERA`  in the manifest, but do not require the user to approve the `Camera` runtime permission - making the calls used in `onCreate` to determine the toggle button safe to perform prior to validating runtime permissions.


samtstern:
@VarPete thanks for this!  LGTM

PR #799, opened 2019-02-01
Use kotlin extensions for langid sample and remove unneeded lib dep i…

gkaldev:
…n mlkit sample.

### DIFF
--- a/mlkit-langid/app/src/main/java/com/google/firebase/samples/apps/mlkit/languageid/kotlin/MainActivity.kt
+++ b/mlkit-langid/app/src/main/java/com/google/firebase/samples/apps/mlkit/languageid/kotlin/MainActivity.kt
@@ -1,17 +1,14 @@
 package com.google.firebase.samples.apps.mlkit.languageid.kotlin
 
 import android.os.Bundle
-import android.support.design.widget.TextInputEditText
 import android.support.v7.app.AppCompatActivity
 import android.text.TextUtils
 import android.util.Log
-import android.widget.Button
 import android.widget.TextView
 import android.widget.Toast
-
 import com.google.firebase.ml.naturallanguage.FirebaseNaturalLanguage
 import com.google.firebase.samples.apps.mlkit.languageid.R
-
+import kotlinx.android.synthetic.main.activity_main.*
 import java.util.ArrayList
 import java.util.Locale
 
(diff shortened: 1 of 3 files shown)


mlkit-langid/app/src/main/java/com/google/firebase/samples/apps/mlkit/languageid/kotlin/MainActivity.kt
```diff
@@ -1,17 +1,14 @@
...
 import android.text.TextUtils
 import android.util.Log
-import android.widget.Button
 import android.widget.TextView
 import android.widget.Toast
-
 import com.google.firebase.ml.naturallanguage.FirebaseNaturalLanguage
 import com.google.firebase.samples.apps.mlkit.languageid.R
-
+import kotlinx.android.synthetic.main.activity_main.*
```
samtstern: If `ktlint` complains about the wildcard imports you can turn it off in Android Studio.  But if it doesn't then this is fine with me.

mlkit-langid/app/src/main/java/com/google/firebase/samples/apps/mlkit/languageid/kotlin/MainActivity.kt
```diff
@@ -22,23 +19,20 @@ class MainActivity : AppCompatActivity() {
...
         setContentView(R.layout.activity_main)
 
-        val inputText = findViewById<TextInputEditText>(R.id.input_text)
-        val idLanguageButton = findViewById<Button>(R.id.button_id_language)
-        val findAllButton = findViewById<Button>(R.id.button_id_all)
         outputText = findViewById(R.id.output_text)
 
-        idLanguageButton.setOnClickListener {
-            val input = inputText.text?.toString()
+        button_id_language.setOnClickListener {
```
samtstern: nit: Could you change the IDs in the XML to be camel case?  So this would look like either `buttonIdLanguage.setOnClickListener` or `idLanguageButton.setOnClickListener`

mlkit-langid/app/src/main/java/com/google/firebase/samples/apps/mlkit/languageid/kotlin/MainActivity.kt
gkaldev: I don't think ktlint complained, but I prefer not having wildcards -- fixed it.

mlkit-langid/app/src/main/java/com/google/firebase/samples/apps/mlkit/languageid/kotlin/MainActivity.kt
gkaldev: Done

Issue #590, opened 2018-08-09
Quick Start Example Cloud Messaging onMessageReceived never called


mrkhanakia:
### Step 1: Are you in the right place?

  * For issues or feature requests related to the code **in this repository** file a Github issue.
    * If you are filing an issue about **FCM in the background** make sure to read [#4](https://github.com/firebase/quickstart-android/issues/4) and [#89](https://github.com/firebase/quickstart-android/issues/89) first!
  * For general technical questions, post a question on [StackOverflow](http://stackoverflow.com/) tagged appropriately.
  * For general Firebase discussion, use the [firebase-talk google group](https://groups.google.com/forum/#!forum/firebase-talk)
  * For help troubleshooting your application that does not fall under one of the above categories, reach out to the [personalized Firebase support channel](https://firebase.google.com/support/contact/troubleshooting/)

### Step 2: Describe your environment

  * Android device: Emulator Nexux
  * Android OS version: Oreo 8
  * Google Play Services version: 4.0.2
  * Firebase/Play Services SDK version: 16.0.1
  
### Step 3: Describe the problem:
I am using QUICK Start MESSAGING example everything is working as excepted but custom sendNotification function is not firing and it is not showing anything.

https://github.com/firebase/quickstart-android/blob/master/messaging/app/src/main/java/com/google/firebase/quickstart/fcm/MyFirebaseMessagingService.java
#### Steps to reproduce:
Just use the quick start example
#### Observed Results:

  * What happened?  This could be a description, `logcat` output, etc.
  
#### Expected Results:

It should log the output as mentioned in the quick start example
  
#### Relevant Code:

  ```
  public void onMessageReceived(RemoteMessage remoteMessage) {
        // [START_EXCLUDE]
        // There are two types of messages data messages and notification messages. Data messages are handled
        // here in onMessageReceived whether the app is in the foreground or background. Data messages are the type
        // traditionally used with GCM. Notification messages are only received here in onMessageReceived when the app
        // is in the foreground. When the app is in the background an automatically generated notification is displayed.
        // When the user taps on the notification they are returned to the app. Messages containing both notification
        // and data payloads are treated as notification messages. The Firebase console always sends notification
        // messages. For more see: https://firebase.google.com/docs/cloud-messaging/concept-options
        // [END_EXCLUDE]

        // TODO(developer): Handle FCM messages here.
        // Not getting messages here? See why this may be: https://goo.gl/39bRNJ
[...]

PR #812, opened 2019-02-23
Only one messaging service at a time

kroikie:
resolves: #590 

Since only one service can be started at a time, having both kotlin and java services for incoming messages will result in only on being launched. To avoid this uncertainty devs should comment one of the services to ensure that the other (service of interest) is started.

### DIFF
--- a/messaging/README.md
+++ b/messaging/README.md
@@ -16,6 +16,9 @@ Getting Started
 
 - [Add Firebase to your Android Project](https://firebase.google.com/docs/android/setup).
 - Run the sample on Android device or emulator.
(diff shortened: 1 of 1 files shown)


samtstern:
Whoops sorry Arthur, I think I beat you to this here:
https://github.com/firebase/quickstart-android/commit/af0ce4d301dcaa877ac33b654ef62e473989c963

(And a correct to my hasty commit is in here https://github.com/firebase/quickstart-android/pull/811)

kroikie:
Cool, thanks!

PR #850, opened 2019-03-26
Make FirebaseMessagingServices not exported

gsakakihara:
* Set FirebaseMessagingServices to exported="false" to explicitly
prevent other apps from being able to send messages to it.
* Fixed documentation to indicate to remove the Kotlin service
intent filter for Java now that the Kotlin service is first in the
manifest.


samtstern:
@gsakakihara thanks Greg!

PR #883, opened 2019-06-07
Convert Quickstart to AndroidX

samtstern:

PR #888, opened 2019-06-12
Update Storage Quickstart to AndroidX

schmidt-sebastian:
Apart from the update to storage v18, the contents here are all auto-generated. The app builds and runs.

This is obviously blocked on the public v18 release.


samtstern:
Being replaced by #883

PR #889, opened 2019-06-12
Update Database to AndroidX

schmidt-sebastian:
This was mostly done by Android Studio apart from:
- Updating all dependencies to their latest version.
- Removing some unnecessary qualified imports that the migration created.
- Removing one unused import in Kotlin


samtstern:
Being replaced by #883

PR #948, opened 2019-09-14
Update dependencies

rosariopfernandes:
The automatic build is failing on #945 because the recent update of appcompat changed some parameters from nullable to non-nullable and vice-versa.
This PR should fix the build.


samtstern:
@rosariopfernandes thank you!  This is great.

PR #956, opened 2019-10-02
Some deprecated code, order code and safe null fix

carlosgub:
Hi I fix some safe Null, order some code and change deprecated code. I will do some more fixes.
Firebase Lover <3


samtstern:
@carlosgub LGTM thanks for the contribution!

PR #990, opened 2019-11-06
Add barcode detection model.

maehara08:
We couldn't use `Barcode Detection` sample.
I added dependency of `firebase-ml-vision-barcode-model` v16.0.1 in this Pull Request


samtstern:
@maehara08 LGTM thank you for the contribution!

PR #995, opened 2019-11-11
Blocked on setStickers call

hborders:
`AppIndexingService`'s call to `AppIndexingUtil#setStickers` is asynchronous, but `IntentService#onHandleIntent` is a synchronous method, so we must return the `Task<Void>` from `AppIndexingUtil#setStickers` and block `onHandleIntent` until the `setStickers` `Task<Void>` completes.


hborders:
Attached with the wrong email.

PR #996, opened 2019-11-11
Blocked on setStickers call

hborders:
`AppIndexingService`'s call to `AppIndexingUtil#setStickers` is asynchronous, but `IntentService#onHandleIntent` is a synchronous method, so we must return the `Task<Void>` from `AppIndexingUtil#setStickers` and block `onHandleIntent` until the `setStickers` `Task<Void>` completes.

### DIFF
--- a/app-indexing/app/src/main/java/com/google/samples/quickstart/appindexing/java/AppIndexingService.java
+++ b/app-indexing/app/src/main/java/com/google/samples/quickstart/appindexing/java/AppIndexingService.java
@@ -17,7 +17,7 @@ public class AppIndexingService extends IntentService {
     @Override
     protected void onHandleIntent(Intent intent) {
         final Task<Void> setStickersTask = AppIndexingUtil.setStickers(getApplicationContext(), FirebaseAppIndex.getInstance());
-        if (task != null) {
+        if (setStickersTask != null) {
            try {
                Tasks.await(setStickersTask); 
            } catch (ExecutionException e) {
--- a/app-indexing/app/src/main/java/com/google/samples/quickstart/appindexing/java/AppIndexingService.java
+++ b/app-indexing/app/src/main/java/com/google/samples/quickstart/appindexing/java/AppIndexingService.java
@@ -8,6 +8,8 @@ import com.google.android.gms.tasks.Tasks;
 
 import com.google.firebase.appindexing.FirebaseAppIndex;
 
+import java.util.concurrent.ExecutionException;
+
 public class AppIndexingService extends IntentService {
 
     public AppIndexingService() {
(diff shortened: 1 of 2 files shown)


app-indexing/app/src/main/java/com/google/samples/quickstart/appindexing/java/AppIndexingUtil.java
```diff
@@ -85,8 +85,11 @@ public void onFailure(@NonNull Exception e) {
                             .show();
                 }
             });
+            
+            return task;
         } catch (IOException | FirebaseAppIndexingInvalidArgumentException e) {
             Log.e(TAG, "Unable to set stickers", e);
+            return null;
```
samtstern: You could return `Tasks.forException(e)` here if you want to avoid ever returning null.

samtstern: Although I can see how this would make your `await` more difficult.

app-indexing/app/src/main/java/com/google/samples/quickstart/appindexing/java/AppIndexingService.java
```diff
@@ -13,6 +16,15 @@ public AppIndexingService() {
 
     @Override
     protected void onHandleIntent(Intent intent) {
-        AppIndexingUtil.setStickers(getApplicationContext(), FirebaseAppIndex.getInstance());
+        final Task<Void> setStickersTask = AppIndexingUtil.setStickers(getApplicationContext(), FirebaseAppIndex.getInstance());
+        if (task != null) {
```
samtstern: This should be `setStickersTask != null`

hborders: Fixed. Please squash-and-merge so that my error doesn't clutter your history.

If you'd prefer, I can squash my change myself.

samtstern:
@hborders ok looks like there's one more issue:
```
> Task :app-indexing:app:compileDebugJavaWithJavac
/home/travis/build/firebase/quickstart-android/app-indexing/app/src/main/java/com/google/samples/quickstart/appindexing/java/AppIndexingService.java:23: error: cannot find symbol
           } catch (ExecutionException e) {
                    ^
  symbol:   class ExecutionException
  location: class AppIndexingService
1 error
> Task :app-indexing:app:compileDebugJavaWithJavac FAILED
```

hborders:
Fixed. Thanks again for your patience. I thought I could do this all from `vim` without setting up a real environment. :)

samtstern:
@hborders thank you for doing this!

PR #1005, opened 2019-11-26
refactor: make kotlin improvements where possible

rosariopfernandes:
Sending some minor kotlin improvements I found while navigating our quickstarts.


samtstern:
@rosariopfernandes looks good to me (as always!)

PR #1010, opened 2019-12-06
refactor(auth): replace ProgressDialog with ProgressBar

rosariopfernandes:
[ProgressDialog](https://developer.android.com/reference/android/app/ProgressDialog) has been deprecated in API 26. This PR should replace its usage on the auth quickstart with a ProgressBar.

Here's how it looks now:
Screenshot from 2019-12-06 13-29-46
Screenshot from 2019-12-06 13-29-00

I'll also send follow-up PRs to replace it in:
- [ ] RTDB Quickstart
- [ ] Storage Quickstart

### DIFF
--- a/auth/app/src/main/java/com/google/firebase/quickstart/auth/java/EmailPasswordActivity.java
+++ b/auth/app/src/main/java/com/google/firebase/quickstart/auth/java/EmailPasswordActivity.java
@@ -57,7 +57,7 @@ public class EmailPasswordActivity extends BaseActivity implements
         mDetailTextView = findViewById(R.id.detail);
         mEmailField = findViewById(R.id.fieldEmail);
         mPasswordField = findViewById(R.id.fieldPassword);
-        mProgressBar = findViewById(R.id.progressBar);
+        setProgressBar(R.id.progressBar);
 
         // Buttons
         findViewById(R.id.emailSignInButton).setOnClickListener(this);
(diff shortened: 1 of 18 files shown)


samtstern:
@rosariopfernandes thanks for starting this!  Could we move the `ProgressBar` so it's at the top of the screen?  I think that's the most common pattern I have seen for a loading bar.

auth/app/src/main/java/com/google/firebase/quickstart/auth/java/EmailPasswordActivity.java
```diff
@@ -57,6 +57,7 @@ public void onCreate(Bundle savedInstanceState) {
         mDetailTextView = findViewById(R.id.detail);
         mEmailField = findViewById(R.id.fieldEmail);
         mPasswordField = findViewById(R.id.fieldPassword);
+        mProgressBar = findViewById(R.id.progressBar);
```
samtstern: Could we move this assignment into `BaseActivity`?  We could either assume that the progress bar is always `R.id.progressBar` or we could introduce a method for the subclasses like `setProgressBar(R.id.progressBar)`

rosariopfernandes:
@samtstern done! Here's how it looks now:

Screenshot from 2019-12-07 15-40-53

samtstern:
@rosariopfernandes thanks for the quick changes!

PR #1018, opened 2019-12-10
refactor(rtdb): replace ProgressDialog with ProgressBar

rosariopfernandes:
Follow up to #1010 
Replacing deprecated `ProgressDialog` with `ProgressBar`.


samtstern:
:+1:

PR #1041, opened 2020-01-12
Only one messaging service at a time

oQuantumDecayo:


samtstern:
@oQuantumDecayo I believe this information is already in the README here:
https://github.com/firebase/quickstart-android/blob/master/messaging/README.md#getting-started

PR #1042, opened 2020-01-13
QuantumDecay Games Android

oQuantumDecayo:


samtstern:
@oQuantumDecayo we would like to move to GitHub actions eventually but this `android.yml` build script is not complex enough to replicate what we were doing in `.travis.yml`.  Do you want to try and update it?

PR #1046, opened 2020-01-15
refactor(analytics): use androidx preference library

rosariopfernandes:
As mentioned in the [documentation](https://developer.android.com/reference/android/preference/PreferenceManager), `android.preference.PreferenceManager` has been deprecated in favor of `androidx.preference.PreferenceManager`. This PR should address that in our analytics sample app.


samtstern:
LGTM thanks for noticing this

PR #1050, opened 2020-01-21
refactor(firestore): Html --> HtmlCompat

rosariopfernandes:
Another one on the deprecation saga. 

This PR should replace the deprecated [`Html.fromHtml()`](https://developer.android.com/reference/android/text/Html.html#fromHtml(java.lang.String)) method with `HtmlCompat.fromHtml()`, as recommended on this [post](https://stackoverflow.com/a/37905107/5861618).


samtstern:
So many new compats!  Thanks for being on top of this, as always.

PR #1056, opened 2020-02-03
Remove memory leak from CustomImageClassifier.kt

FuadRafid:
The imgData variable in convertBitmapToByteBuffer function was being assigned a new byte[] each time, and the last used byte[] were not being removed from memory which was causing a memory leak. The variable was made a property of the class and was initiated only once, cleared and reused again.


FuadRafid:
@googlebot I signed it!

samtstern:
@FuadRafid this LGTM thank you for the contribution!

PR #1062, opened 2020-02-13
Create android.yml

DarthGrave:


samtstern:
@DarthGrave as you can see in our `.travis.yml` this repository has a pretty complex CI build setup and this `android.yml` file will not replicate that.  This looks like just a standard GitHub Actions template file.

Since this PR has no justification I assume it was a mistake.

PR #1083, opened 2020-03-13
Multi Factor Authentication

samtstern:
TODO:

- [x] README
- [x] Disclaimer
- [x] If you enroll a second factor and try to sign-in with email/password, it just fails with "Authentication failed" - does not go to the MultiFactorSignInActivity


samtstern:
cc @morganchen12

auth/app/build.gradle
```diff
@@ -40,7 +40,8 @@ dependencies {
     implementation 'com.google.android.material:material:1.1.0'
 
     // Firebase Authentication
-    implementation 'com.google.firebase:firebase-auth:19.2.0'
+    implementation 'com.google.firebase:firebase-core:17.2.3'
```
rosariopfernandes: Shouldn't we use the analytics library instead?

```suggestion
    implementation 'com.google.firebase:firebase-analytics:17.2.3'
```

samtstern: Good catch!

PR #1085, opened 2020-03-18
test: replace deprecated AndroidJUnit4

rosariopfernandes:
`androidx.test.runner.AndroidJUnit4` is now deprecated in favor of `androidx.test.ext.junit.runners.AndroidJUnit4`. This PR should update our tests to use the new class instead.

Additionally, it should replace the deprecated functions `Espresso.registerIdlingResources()` and `Espresso.unregisterIdlingResources()` with `IdlingRegistry.getInstance().register()` and `IdlingRegistry.getInstance().unregister()` respectively.


samtstern:
@rosariopfernandes as always, thank you!

PR #1094, opened 2020-04-07
chore: implement selective builds on Github Actions

rosariopfernandes:
Now this is just an idea I got while reading [this post](https://overflow.buffer.com/2019/12/20/selectively-running-android-modularized-unit-tests-on-your-ci-server/) by Joe Birch.

The config created in this PR should run our build selectively, which means it'll only build the modules that have been changed in a PR.
The flow goes as following:
- If the build was triggered by a push, the whole project will be built using `./gradlew clean ktlint build` (no difference from our current flow)
- If the build was triggered by a PR:
    1. use `git diff` to check which files were changed;
    1. check which of these files belong to a gradle module;
    1. run `./gradlew clean ktlint`
    1. build the changed modules using `:module:assembleDebug` and `:module:check`. (If no modules were changed, this step is skipped)

This results in faster build times for our PRs.

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -17,5 +17,11 @@ jobs:
       run: echo "y" | sudo ${ANDROID_HOME}/tools/bin/sdkmanager --install "ndk;20.0.5594570" --sdk_root=${ANDROID_SDK_ROOT}
     - name: Check Snippets
       run: python scripts/checksnippets.py
-    - name: Build with Gradle
-      run: ./build.sh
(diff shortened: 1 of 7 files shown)


samtstern:
@rosariopfernandes this is super cool, thank you for doing it!

rosariopfernandes:
@samtstern It is always a pleasure. Also, do you think it's worthy copying this config over to our snippets repo?

samtstern:
Yeah I do think that would be useful

PR #1101, opened 2020-04-12
Update README.md

enigmaharjan:
The name of the application was incorrect


samtstern:
@enigmaharjan thanks!  Fire Eats was the name of the app wayyyy back in 2017 when we were testing internally.

PR #1109, opened 2020-04-23
Fix ViewBinding build

samtstern:
Small issue in MLKit

cc @rosariopfernandes


samtstern:
@rosariopfernandes yeah I did the same for some of the simple ones but I like ViewBinding so much now I want to use it everywhere!

Crazy how long it took for Android to get this simple API.

rosariopfernandes:
@samtstern Hehe! In case you didn't know, there's also [DataBinding](https://developer.android.com/topic/libraries/data-binding), which is very similar to View Binding, but the main difference is that it supports 2-way binding.

samtstern:
@rosariopfernandes yeah I never liked Data Binding because it was always kinda fragile but it's probably gotten a lot better so I will take a look again.

PR #1110, opened 2020-04-24
chore: fix pull request build

rosariopfernandes:
On #1096, Github Action didn't build mlkit, but did build mlkit-langid and mlkit-smartreply. mlkit was not being added to the list of modules because the list already contains "`*mlkit*`" (`*` means any). This should update our script to match `*mlkit` instead.

I tested this at my fork: rosariopfernandes/quickstart-android#8


samtstern:
@rosariopfernandes ah that explains it!

PR #1114, opened 2020-04-27
refactor(auth): change firebaseAuthWithGoogle() argument type

rosariopfernandes:
The main reason why I've opened this PR is to change the `firebaseAuthWithGoogle()` function to take a googleIdToken (`String`) as argument instead of `GoogleSignInAccount`. Since some Auth libraries like [One Tap UI](https://developers.google.com/identity/one-tap/android/get-saved-credentials#sign_in_with_retrieved_credentials) actually return google id tokens, we could reuse the `firebaseAuthWithGoogle()` method to handle both: a simple google sign in and a google sign with One Tap UI.

---
Additionally, I've found a few smaller issues and I fixed them:
- some Kotlin function declarations had `public` on their signature
- the `binding` property was exposed in a snippet


samtstern:
@rosariopfernandes this is a good simplification, thank you!

PR #1116, opened 2020-05-04
refactor(analytics): update code to use Analytics KTX

rosariopfernandes:


samtstern:
@rosariopfernandes looks goodd!

PR #1134, opened 2020-06-18
fix(messaging): fix wrong nullability check on MessagingService

rosariopfernandes:
The way it's currently written, the `let` block will always be called regardless of whether `isNotEmpty` returns `true` or `false`.

Update: I also noticed we could've used a string template instead of concatenation on the same method. I pushed a new commit changing it.


samtstern:
Nice catch

PR #1148, opened 2020-07-24
Create SECURITY.md

Miu98:


samtstern:
@Miu98 I don't think this file is relevant to this quickstart repo.  We don't have "versions" that are supported.

PR #1149, opened 2020-07-24
Crashlytics legacy

Miu98:


samtstern:
Looks like this was a mistake.

PR #1168, opened 2020-09-03
refactor(storage): use the new Kotlin Destructuring Declarations

rosariopfernandes:
This PR should update our storage sample app to use the Kotlin Destructuring Declarations introduced in `storage-ktx:19.2.0`.

It also fixes a couple of warnings on `MyUploadService.kt`:
- The `Service` import is redundant.
- `Log.d()` can use String template instead of String concatenation.


samtstern:
@rosariopfernandes I missed this one too!

PR #1169, opened 2020-09-10
chore(*): remove kotlin-stdlib-jdk7 dependency

rosariopfernandes:
Starting in Kotlin 1.4.0, the kotlin-stdlib is [added by default](https://kotlinlang.org/docs/reference/whatsnew14.html#dependency-on-the-standard-library-added-by-default). So we no longer need it on our build.gradle files.


samtstern:
@rosariopfernandes nice!  CI passes to LGTM

PR #1191, opened 2020-10-09
Add some more KTX libs, use Installations instead of InstanceId

samtstern:
Changes

- Add Perf KTX
- Add Messaging KTX
- Remove uses of FirebaseInstanceId

### DIFF
--- a/inappmessaging/app/src/main/java/com/google/firebase/fiamquickstart/kotlin/KotlinMainActivity.kt
+++ b/inappmessaging/app/src/main/java/com/google/firebase/fiamquickstart/kotlin/KotlinMainActivity.kt
@@ -8,9 +8,9 @@ import com.google.firebase.analytics.FirebaseAnalytics
 import com.google.firebase.analytics.ktx.analytics
 import com.google.firebase.fiamquickstart.R
 import com.google.firebase.fiamquickstart.databinding.ActivityMainBinding
-import com.google.firebase.iid.FirebaseInstanceId
 import com.google.firebase.inappmessaging.FirebaseInAppMessaging
 import com.google.firebase.inappmessaging.ktx.inAppMessaging
+import com.google.firebase.installations.FirebaseInstallations
 import com.google.firebase.ktx.Firebase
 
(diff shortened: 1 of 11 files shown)


inappmessaging/app/src/main/java/com/google/firebase/fiamquickstart/kotlin/KotlinMainActivity.kt
```diff
@@ -36,12 +36,12 @@ class KotlinMainActivity : AppCompatActivity() {
...
         }
 
-        // Get and display/log the Instance ID
-        FirebaseInstanceId.getInstance().instanceId
-                .addOnSuccessListener { instanceIdResult ->
-                    val instanceId = instanceIdResult.id
-                    binding.instanceIdText.text = getString(R.string.instance_id_fmt, instanceId)
-                    Log.d(TAG, "InstanceId: $instanceId")
+        // Get and display/log the installation id
+        FirebaseInstallations.getInstance()
```
rosariopfernandes, line 40: Since we're using KTX, I think we could use ` Firebase.installations` instead

messaging/app/src/main/java/com/google/firebase/quickstart/fcm/kotlin/MainActivity.kt
```diff
@@ -64,21 +64,22 @@ class MainActivity : AppCompatActivity() {
         binding.logTokenButton.setOnClickListener {
             // Get token
             // [START retrieve_current_token]
-            FirebaseInstanceId.getInstance().instanceId
-                    .addOnCompleteListener(OnCompleteListener { task ->
+            FirebaseInstallations.getInstance()
```
rosariopfernandes, line 67: Here too

inappmessaging/app/src/main/java/com/google/firebase/fiamquickstart/kotlin/KotlinMainActivity.kt
samtstern: Ah cool I didn't know installations had KTX too

PR #1208, opened 2020-11-05
refactor(auth): fix kotlin errors and improve chooser layout

rosariopfernandes:
Fixing 2 errors here:
- `setContentView()` was being called with the wrong argument in ChooserActivity, causing the `ListView` to not be shown
- On MultiFactorActivity, `setContentView()` was being called with an incompatible `View`, causing the app to crash.

This PR should also remove the padding around the ListView in ChooserActivity and replace its `LinearLayout` with a `FrameLayout`.


samtstern:
@rosariopfernandes thanks for catching these mistakes!

samtstern (approved):

PR #1215, opened 2020-11-19
docs(functions): fix StackOverflow url in README.md

rosariopfernandes:
The StackOverflow URL in our functions README.md was incorrect.


samtstern:
👍  nice catch!

PR #1220, opened 2020-11-26
refactor(database, firestore): remove kotlin synthetic binding

rosariopfernandes:
Seems like we missed 2 quickstarts when removing Kotlin Synthetic Binding, which is [now deprecated](https://android-developers.googleblog.com/2020/11/the-future-of-kotlin-android-extensions.html).

### DIFF
--- a/database/app/build.gradle
+++ b/database/app/build.gradle
@@ -1,6 +1,5 @@
 apply plugin: 'com.android.application'
 apply plugin: 'kotlin-android'
-apply plugin: 'org.jetbrains.kotlin.android.extensions'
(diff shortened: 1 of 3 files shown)


samtstern:
@rosariopfernandes huh how did we miss these two?  I totally remember doing the Firestore one ... maybe I never pushed it.

rosariopfernandes:
@samtstern I think you did push it. But these 2 were using the Experimental version of the `kotlin-android-extensions` plugin, which I guess you didn't remove because that version is applied differently.

PR #1222, opened 2020-12-03
refactor(admob): replace RelativeLayout with ConstraintLayout

rosariopfernandes:
Kick starting the "jetpackification" with a simple one.

Here are the new layouts:

activity_main.xml
Screenshot from 2020-12-04 00-44-08

second_activity.xml
Screenshot from 2020-12-04 00-44-25

### DIFF
--- a/admob/app/src/main/res/layout/activity_main.xml
+++ b/admob/app/src/main/res/layout/activity_main.xml
@@ -1,29 +1,32 @@
 <!-- [SNIPPET define_ad_layout]
 Define the namespace and AdView layout that will show the banner.
 [START define_ad_layout] -->
-<RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
+<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:ads="http://schemas.android.com/apk/res-auto"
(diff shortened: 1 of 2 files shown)


admob/app/src/main/res/layout/activity_main.xml
```diff
@@ -1,29 +1,32 @@
...
     tools:context=".java.MainActivity">
 
     <ImageView
         android:layout_width="wrap_content"
         android:layout_height="wrap_content"
-        android:layout_centerHorizontal="true"
-        android:src="@drawable/firebase_lockup_400"/>
+        android:layout_marginTop="16dp"
+        android:src="@drawable/firebase_lockup_400"
+        ads:layout_constraintEnd_toEndOf="parent"
```
samtstern, line 17: Should these be `app:` not `ads:`?

samtstern: Same question for all the `ads:` properties that deal with constraints.

rosariopfernandes: @samtstern You're right. I changed this through the layout editor, so I guess Android Studio picked that namespace instead of app by mistake.

Fixed in the latest commit.

samtstern (approved):

PR #1224, opened 2020-12-04
Josephwambura patch binding bugs

josephwambura:
A few fixes, first, it is some useless texts in the strings file (firestore project) and a bug found in the FilterDialog

### DIFF
--- a/firestore/app/src/main/java/com/google/firebase/example/fireeats/kotlin/FilterDialogFragment.kt
+++ b/firestore/app/src/main/java/com/google/firebase/example/fireeats/kotlin/FilterDialogFragment.kt
@@ -144,7 +144,7 @@ class FilterDialogFragment : DialogFragment() {
     }
 
     fun resetFilters() {
-        binding.let {
+        binding?.let {
             it.spinnerCategory.setSelection(0)
             it.spinnerCity.setSelection(0)
             it.spinnerPrice.setSelection(0)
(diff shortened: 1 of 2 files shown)


josephwambura:
@googlebot I signed it!

firestore/app/src/main/java/com/google/firebase/example/fireeats/kotlin/FilterDialogFragment.kt
```diff
@@ -144,7 +144,7 @@ class FilterDialogFragment : DialogFragment() {
     }
 
     fun resetFilters() {
-        binding.let {
+        binding?.let {
```
rosariopfernandes, line 147: In this case, `binding` is never null. In order to fix the error, we might want to use `_binding` instead.
```suggestion
        _binding?.let {
```

firestore/app/src/main/res/values/strings.xml
```diff
@@ -61,7 +61,6 @@
         <item>Ramen</item>
         <item>Sushi</item>
     </string-array>
-    ite
```
rosariopfernandes, line 64: Good catch!

rosariopfernandes (changes requested):

firestore/app/src/main/java/com/google/firebase/example/fireeats/kotlin/FilterDialogFragment.kt
josephwambura: Crashed if i used the implementation in a fragment, calling the Dialog from a fragment. But yeah _binding?.let is better

firestore/app/src/main/res/values/strings.xml
josephwambura: Nice one

rosariopfernandes (approved):

samtstern:
@josephwambura and @rosariopfernandes thank you both!

PR #1241, opened 2021-01-03
Updated comments for more clarity on when onNewToken is called

joeykrim:
Given the lengthy history of obtaining the token from C2DM, GCM, FCM, and the transition from InstanceID to Installation SDK, added additional clarity to outline all the known scenarios onNewToken is called to match with the details from https://firebase.google.com/docs/cloud-messaging/android/client#sample-register
Attempting to prevent reading multiple blog posts and docs to understand when to use the method in the sample and rather incorporate the latest known information into the sample by updating the previous comments to the latest information


joeykrim:
Will resubmit under different email address to match CLA

PR #1237, opened 2020-12-31
refactor(auth): remove RelativeLayout usage

rosariopfernandes:
See discussion in #1217 

**~This is a Work-In-Progress:~**
### Anonymous Auth
<table><thead><tr><th>Before</th><th>After</th></tr></thead><tbody><tr><td>

</td><td>

</td></tr></tbody></table>

### Email Password
<table><thead><tr><th>Before</th><th>After</th></tr></thead><tbody><tr><td>

</td><td>

</td></tr></tbody></table>

### Facebook
<table><thead><tr><th>Before</th><th>After</th></tr></thead><tbody><tr><td>

</td><td>

</td></tr></tbody></table>

### Firebase UI
<table><thead><tr><th>Before</th><th>After</th></tr></thead><tbody><tr><td>

</td><td>

</td></tr></tbody></table>

### Generic IDP
<table><thead><tr><th>Before</th><th>After</th></tr></thead><tbody><tr><td>

</td><td>

</td></tr></tbody></table>

### Google
<table><thead><tr><th>Before</th><th>After</th></tr></thead><tbody><tr><td>

</td><td>

</td></tr></tbody></table>

### Multi Factor Auth

I changed this one to give it a better look because the buttons were all squeezed in there.

<table><thead><tr><th>Before</th><th>After</th></tr></thead><tbody><tr><td>

</td><td>

</td></tr></tbody></table>

### Passwordless
<table><thead><tr><th>Before</th><th>After</th></tr></thead><tbody><tr><td>

</td><td>

</td></tr></tbody></table>

### Phone Auth
<table><thead><tr><th>Before</th><th>After</th></tr></thead><tbody><tr><td>

</td><td>

</td></tr></tbody></table>

Issue #1217, opened 2020-11-20
Jetpacktify our quickstarts ✨


rosariopfernandes:
Following up on my suggestion to implement [Modern Android App Architecture](https://developer.android.com/jetpack/guide) here, I've come up with a list of actionable tasks which we can discuss in this issue.

### 1. Replace RelativeLayout with ConstraintLayout

[ConstraintLayout](https://developer.android.com/training/constraint-layout) seems to be the recommended way to design layouts in Android, right now. I suggest we replace all our `RelativeLayout` usages with `ConstraintLayout`.

#### Quickstarts to be updated

- [ ] Admob
- [ ] App Indexing
- [ ] Authentication
- [ ] Remote Config
- [ ] Crashlytics
- [ ] Realtime Database
- [ ] Dynamic Links
- [ ] Cloud Firestore
- [ ] Cloud Functions
- [ ] Cloud Storage

### 2. Use the Navigation Component

Starting in Android Studio 4.0, creating a new project now gives you a single activity with 2 fragments and the [Navigation Component](https://developer.android.com/guide/navigation) included to help navigate between the 2 fragments.
I suggest we update our quickstarts to use the same single-activity-multiple-fragments approach with the Navigation Component.

#### Quickstarts to be updated

- [ ] Admob
- [ ] Authentication
- [ ] Realtime Database
- [ ] Cloud Firestore

### 3. Use ViewModel+LiveData+Repository

The [Android's App Architecture Guide](https://developer.android.com/jetpack/guide#recommended-app-arch) recommends structuring the app with a ViewModel and a Repository for better separation of concerns.
Not only will this be a huge refactor, but we might also run into other issues because of the quickstarts which include snippets that are shown on the Firebase Docs. So I'd suggest we start with the databases quickstarts first, to see if it's worth refactoring the others.

- [ ] Cloud Firestore
- [ ] Realtime Database

samtstern:
@rosariopfernandes thanks for starting the conversation and breaking down the tasks!  

- I think (1) is an obvious positive, we should do it.
- I also think (2) is probably a good idea, I don't think it adds much complexity
- I am a bit concerned about (3) ... I'll have to see what this looks like in an example before deciding.  Quickstarts need to be careful not to "hide" the firebase behind platform abstractions

Oh and don't worry about snippets. I'd be happy to have this excuse to transfer more of them to `snippets-android` anyway.

rosariopfernandes:
[...]

PR #1246, opened 2021-01-13
refactor(auth, analytics): remove RelativeLayout from xml files

rosariopfernandes:
Part of #1217 
Seems like I had missed a layout when working on #1237 and discovered analytics was also using `RelativeLayout`.

<table><thead><tr><th>Before</th><th>After</th></tr></thead><tbody><tr><td>

</td><td>

</td></tr></tbody></table>


samtstern:
Thanks for cleaning this up!

PR #1247, opened 2021-01-13
refactor(admob): use Jetpack's Navigation Component

rosariopfernandes:
Kicking Off step 2 of #1217 : Navigation Component ✨ 

This PR should replace the 2 activities in Admob with 2 fragments (`FirstFragment` and `SecondFragment`) and link them using the Navigation Component.

The navigation graph is simple:

Screenshot from 2021-01-13 23-34-48

~Some code changes that I should point out:~
~1. There were snippets in the `MainActivity.java` which were moved to `FirstFragment.java`~
~2. Snippets in `MainActivity.kt` were moved to `FirstFragment.kt`~
~3. Snippets in `activity_main.xml` were moved to `fragment_first.xml`~
~4. The method `beginSecondActivity` was renamed to `goToNextFragment`~

We're moving the snippets from this repo to firebase/snippets-android. See firebase/snippets-android#272

---

I was unsure of what to do with the tests, since they make use of Activities.
I tried replacing it with the code I found in [Test your Fragments](https://developer.android.com/guide/fragments/test), but I'm not sure if it even works.


samtstern:
@rosariopfernandes thank you for doing this and for looking out for the snippet changes.  I'll need to do some internal work to make sure we don't break the docs when we submit this, so that will delay things a bit.

rosariopfernandes:
@samtstern I noticed the `snippets-android` has an admob module. What if we move these snippets there? I can open a PR for this change.

samtstern:
@rosariopfernandes yeah eventually I want to get all of the snippets out of this repo and into snippets-android (unless they're special ones like AndroidManifest.xml snippets).

If you're willing to make that PR that would be great!  You can drop anything in a `[START_EXCLUDE]`/`[END_EXCLUDE]` block and just replace it with `// ...` (or nothing) since that code is not shown in the docs anyway.

rosariopfernandes:
@samtstern What's special about AndroidManifest.xml snippets? I think I saw a snippet in the admob's Manifest. I wonder if maybe I shouldn't move it?

One more question: When moving the snippets, should we keep using Activities? Or do we switch to Fragments?
In the admob snippets specifically, the main difference is that instead of passing the `Activity` as context to the AdMob SDK, we use `Fragment.getContext()`.

samtstern:
@rosariopfernandes there's nothing really special about them but they're sometimes hard to "compile" in the snippets repo, having them in a real app makes them more likely to be correct.  If they're tied to the Java code in the snippets repo though then it's totally fine to move them there as well.

As for the Fragment thing I'd suggest just separating out a variable, so if you previously had this in an activity:

```
doAThing(this)
```

Instead we could do:
```
Context context = this.getContext();
doAThing(context)
```

Which should make it clear to the developer that any `Context` is what they need and our use of a `Fragment` in this app is an implementation detail.

rosariopfernandes:
@samtstern Ah, I see. I'll be sure to watch out for that. Thanks

That Context solution sounds great! I'll do that.

samtstern (approved):
This looks great! Just waiting on it to be safe for snippets (internal change pending)

samtstern:
Internal change merged so now I can merge this one as well!

PR #1250, opened 2021-01-21
refactor(admob): create a separate nav graph for the java sample

rosariopfernandes:
I noticed I made a mistake when I sent #1247 : I only added the Kotlin Fragments to the navigation graph, which caused the java ones to never be used.

As a solution (in this PR), I've created a separate nav graph file for the java fragments (`nav_graph_java.xml`) and the graphs are now set programmatically (instead of setting it in the XML layout).


samtstern:
@rosariopfernandes good catch!  By the middle of 2021 we should be able to go Kotlin-only on these samples. I'm waiting for some internal Kotlin adoption charts to hit certain thresholds and then we can get some heavy `rm -rf java` action going :-)

rosariopfernandes:
@samtstern That's some great news :) Can't wait for it

Issue #275, opened 2017-05-26
migrating from GCM and token scope issue


llooz:
I'm migrating an app from gcm sdk to the firebase sdk i have noticed that, by following the documentation and this quickstart, a new token is generated with a "*" scope without invalidating the old gcm token.

The result is that i have 2 valid token for the same device and so i will send a duplicated notification.

Currently i have solved the issue by forcing the GCM scope that is recommended to use on the gcm documentation and that i'm using on the current gcm version of the app:
`FirebaseInstanceId.getInstance().getToken(getString(R.string.gcm_defaultSenderId), "GCM");`

Am i doing something wrong? Is this an expected behavior?

kroikie:
Hi @llooz the way you are handling the multiple valid tokens looks good. That way getToken will return the same token that GCM returned.

You could also consider removing extra tokens that have the same Instance ID, the part of the token before the colon (:). Since that will be the same for the same device whether you use "FCM" or "*" as the scope.

PR #1252, opened 2021-01-28
refactor(database): use Jetpack's Navigation Component

rosariopfernandes:
Part of #1217, this should implement the Navigation Component in our database quickstart.

## Extra changes

In order to accommodate the Navigation implementation, some extra changes took place here:
- deleted the `NewPostTest` unit test class which was testing one of our activities that no longer exists.
- edited the `AppTheme.NoActionBar` theme because it was causing the systembar to disappear.
- the `fragment` package which contains the fragments used in a `TabLayout` was renamed to `listfragments` to avoid any confusion.
- moved the `CommentViewHolder` out of the `PostDetailActivity` class and into a separate class in the `viewholder` package (now sitting pretty alongside its brother `PostViewHolder`).

## Snippets moved to firebase/snippets-android
See #275

@samtstern there are some snippets that are present in this repo, but I can't find them in the docs.
For example, the `my_top_posts_query`, which is similar to another snippet already present in snippets-android:

#### quickstart-android

https://github.com/firebase/quickstart-android/blob/49bcffe31c562d4106d8338c7cc5aeb9d83a70d7/database/app/src/main/java/com/google/firebase/quickstart/database/java/fragment/MyTopPostsFragment.java#L12-L17

#### snippets-android
https://github.com/firebase/snippets-android/blob/45ba4822eb111609be2951c944142e3bbe891472/database/app/src/main/java/com/google/firebase/referencecode/database/QueryActivity.java#L98-L113

I can see its usage in the docs: https://firebase.google.com/docs/database/android/lists-of-data#sort_data

Snippets in a similar situation include: `comment_class`, `single_value_read` and `declare_database_ref`. Is it possible to check these against the snippet registry?


database/app/src/main/java/com/google/firebase/quickstart/database/java/MainActivity.java
```diff
@@ -16,91 +16,58 @@
...
+                        fab.setVisibility(View.VISIBLE);
+                        fab.setOnClickListener(new View.OnClickListener() {
+                            @Override
+                            public void onClick(View view) {
+                                navController.navigate(R.id.action_MainFragment_to_NewPostFragment);
+                            }
+                        });
+                        break;
+                    }
+                }
```
rosariopfernandes, line 69: note to self: this can be replaced with an if-else

samtstern (approved):

samtstern:
@rosariopfernandes thank you *so* much for being so careful about snippets as you do this. That's really helpful.

So here's what we'll do:
1. Merge your `snippets-android` change
2. Make the docs point to the new snippets (internal change)
3. Merge this one.

samtstern:
This one is ready to go!  Internal change is in so this should be safe.

Issue #1253, opened 2021-02-02
our little dpebot got stuck/blocked :(


rosariopfernandes:
@dpetbot seems to be pushing changes to the correct branch (`dpebot-repositorygardener`), but it mentions a PR that doesn't seem to exist: #1231. I tried opening a PR with the changes from that branch, but GitHub doesn't allow it because it detected a similar open PR (our ghost #1231).

samtstern:
@rosariopfernandes huh I have no idea how a ghost PR like #1231 could happen!

russparker37:
@

samtstern:
See #1255

PR #1254, opened 2021-02-02
Auto-update dependencies.

samtstern:
See #1253

After this I will merge master-temp to master


samtstern:
Hmmm actually maybe this won't work though because #1251 seems like it will be a problem forever.  I might ask GitHub support.

samtstern:
GitHub support fixed this, see #1255

PR #1259, opened 2021-02-27
Modifies YAML files to work with studio/sample page

codingjeremy:
Now that the default is 'main' for all branches. I need to change your default branch is master.

I also fixed the MLKit sample which is listed as a separate sample.


samtstern:
LGTM!

PR #1273, opened 2021-03-29
Add increment sample

puf:
Adds code sample of using increment() operation to replace two transactions with a single multi-path update call.

### DIFF
--- a/database/app/src/main/java/com/google/firebase/quickstart/database/java/listfragments/PostListFragment.java
+++ b/database/app/src/main/java/com/google/firebase/quickstart/database/java/listfragments/PostListFragment.java
@@ -21,11 +21,15 @@ import com.google.firebase.database.DatabaseReference;
 import com.google.firebase.database.FirebaseDatabase;
 import com.google.firebase.database.MutableData;
 import com.google.firebase.database.Query;
+import com.google.firebase.database.ServerValue;
 import com.google.firebase.database.Transaction;
 import com.google.firebase.quickstart.database.R;
 import com.google.firebase.quickstart.database.java.PostDetailFragment;
 import com.google.firebase.quickstart.database.java.models.Post;
 import com.google.firebase.quickstart.database.java.viewholder.PostViewHolder;
+
+import java.util.HashMap;
+import java.util.Map;
 
 public abstract class PostListFragment extends Fragment {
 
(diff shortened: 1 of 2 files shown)


database/app/src/main/java/com/google/firebase/quickstart/database/java/listfragments/PostListFragment.java
```diff
@@ -111,13 +115,19 @@ public void onClick(View v) {
...
                     @Override
                     public void onClick(View starView) {
-                        // Need to write to both places the post is stored
-                        DatabaseReference globalPostRef = mDatabase.child("posts").child(postRef.getKey());
-                        DatabaseReference userPostRef = mDatabase.child("user-posts").child(model.uid).child(postRef.getKey());
-
-                        // Run two transactions
-                        onStarClicked(globalPostRef);
-                        onStarClicked(userPostRef);
+                        if (false) {
```
samtstern, line 118: Not really sure about adding an `if (false)` branch to our quickstart ... this code will never run and quickstarts are meant to be runnable.

samtstern (changes requested):

puf:
I just wasn't sure whether it was worth adding a UI element for this.

So at the moment I simply used a boolean toggle.

Your choice, I can either add a UI element, or add a long comment in the
code explaining what the difference is. I'd probably also add my security
rules snippet there.

samtstern:
@puf I guess my question is: does this need to be in the quickstarts at all?  One benefit of separating the snippets and the quickstarts is that the quickstarts don't need to be bent so that they're exhaustive catalogs of the API.  If there are two ways to do this task now we should probably just use the one that's better for this quickstart.

puf:
I mostly wrote the code in the quickstart because I really wanted to run
it, and that has the UI I want in there already.
But fair point indeed, that doesn't mean we have to *keep* it in this
repository.

I'm actually not sure which approach is *better*.
I would typically use an increment/decrement and server-side security rules
nowadays to implement this functionality, especially since the app already
knows whether this user liked this post or not, as it shows that as the
current state for the star.
But I can well imagine that this is a personal preference, as I definitely
tend to bias towards such solutions.

If we decide to keep my newer approach, I'll augment it to also handle
unstarring a post, update the security rules to allow that, and include the
security rules in comments in the code.

samtstern:
@puf that makes sense, either option in this quickstart is totally fine with me!  I think just go with your instinct of how you'd do this if you were starting today.

puf:
Dropping this in favor of https://github.com/firebase/snippets-android/pull/290

PR #1276, opened 2021-04-07
UI: Change AnonymousAuthFragment Layout

EdNgulele:
### Summary

This PR updates the layout of the AnonymousAuthFragment in order to match with the design of the other layouts (e.g: EmailPasswordFragment, GenericIdpFragment, etc)

Submission Checklist

- [x] I read and followed the [contribution guidelines](https://github.com/firebase/quickstart-android/blob/master/CONTRIBUTING.md)
- [x] All tests passed

### Pending tasks
- Improve layout responsiveness to screen rotation. I noticed that most of the layouts don't respond well when the screen is rotated. I would be happy to work to fix this. 

### Before the change: 
Screenshot_20210407-161923_Firebase Authentication

### After the change
Screenshot_20210407-180132_Firebase Authentication


rosariopfernandes (approved):
LGTM.
Thanks for your contribution, @EdNgulele !

rosariopfernandes:
> I noticed that most of the layouts don't respond well when the screen is rotated. I would be happy to work to fix this.

@EdNgulele If it's a matter of changing some properties/constraints in the existing layouts to make them look good in landscape, I think it's doable.

EdNgulele:
Alright @rosariopfernandes, I will be working on this. Thank you!

auth/app/src/main/res/layout/fragment_anonymous_auth.xml
```diff
@@ -1,136 +1,185 @@
...
+
+        <TextView
+            android:id="@+id/anonymousStatusEmail"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            tools:text="Email: none" />
+
+    </LinearLayout>
+
+    <androidx.constraintlayout.widget.ConstraintLayout
```
samtstern, line 59: Is there a reason why you chose to go with a `LinearLayout` as the top container but then have a `ConstraintLayout` nested within?

samtstern (commented):
Thank you for contributing! I have a question about the new choice of layouts, since we prefer to use ConstraintLayout rather than nested layouts when possible.

auth/app/src/main/res/layout/fragment_anonymous_auth.xml
rosariopfernandes: I think most of our layouts are have been created like that :sweat_smile:  :

https://github.com/firebase/quickstart-android/blob/0e5d584cf573f60d988bf8a803c1e45bec49176a/auth/app/src/main/res/layout/fragment_emailpassword.xml#L1-L24

EdNgulele: Yes @rosariopfernandes, I was actually  trying to follow the way the other layouts were created! 👌🏽

@samtstern if you prefer a different approach, I could find a way to minimize nested  layouts (when possible). But this also means doing the same exercise with the other layouts as well. Which I will also be happy to do.

samtstern: @EdNgulele that's ok, we can leave it as-is!

PR #1284, opened 2021-04-25
fix(docs): corrects typo in project documentation

plan-do-break-fix:
Scope of changes is restricted to docs only. Change is in agreement with Standard American English.


samtstern:
@plan-do-break-fix thank you!

PR #1289, opened 2021-05-06
upgrade to AGP 4.2 and update dependencies

rosariopfernandes:
I've pulled the changes from #1285 and fixed the build errors by:
- Updating the gradle version (6.5.1 --> 6.7.1)
- Updating our `internal-lint` module to use Java 8.

### DIFF
--- a/internal/lint/build.gradle
+++ b/internal/lint/build.gradle
@@ -5,10 +5,10 @@ targetCompatibility = JavaVersion.VERSION_1_7
 sourceCompatibility = JavaVersion.VERSION_1_7
 
 dependencies {
-    compileOnly "com.android.tools.lint:lint-api:27.1.3"
-    compileOnly "org.jetbrains.kotlin:kotlin-stdlib-jdk7:1.4.32"
-    testImplementation "com.android.tools.lint:lint:27.1.3"
-    testImplementation "com.android.tools.lint:lint-tests:27.1.3"
+    compileOnly "com.android.tools.lint:lint-api:27.2.0"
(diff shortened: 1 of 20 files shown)


internal/lint/build.gradle
```diff
@@ -1,14 +1,23 @@
...
-targetCompatibility = JavaVersion.VERSION_1_7
-sourceCompatibility = JavaVersion.VERSION_1_7
+repositories {
+    mavenCentral()
+}
+
+java {
+    targetCompatibility = JavaVersion.VERSION_1_8
+    sourceCompatibility = JavaVersion.VERSION_1_8
+}
```
rosariopfernandes, line 11: I wonder if we had a strong reason to pin this at Java 7, but I had to upgrade it because:
1. AGP 4.2 now builds with Java 8 by default;
2. Starting in v27.2.0, the `com.android.tools.lint:lint-api` library only compiles with Java 8. Trying to use the new version with Java 7 results in an error.

internal/lint/build.gradle
```diff
@@ -1,14 +1,23 @@
 apply plugin: 'java-library'
 apply plugin: "kotlin"
 
-targetCompatibility = JavaVersion.VERSION_1_7
-sourceCompatibility = JavaVersion.VERSION_1_7
+repositories {
+    mavenCentral()
+}
```
rosariopfernandes, line 6: I had to specify the maven repo here because gradle couldn't find the `org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.5.0` repository.

PR #1301, opened 2021-06-10
chore (crash): set minifyEnabled to false and update dependencies

rosariopfernandes:
I don't know if we had any strong reason to keep `minifyEnabled` set to `true`, but it was breaking our build in #1299

### DIFF
--- a/admob/app/build.gradle
+++ b/admob/app/build.gradle
@@ -41,7 +41,7 @@ dependencies {
     implementation 'androidx.navigation:navigation-fragment-ktx:2.3.5'
     implementation 'androidx.navigation:navigation-ui-ktx:2.3.5'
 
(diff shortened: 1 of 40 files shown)


samtstern:
@rosariopfernandes I think the build breakage in #1299 is correct. The reason for `minifyEnabled: true` in this repo is to make sure that Firebase SDKs compile with minification enabled. Sort of a last-resort smoke test.

I think the crashlytics plugin has a bug right now with Gradle 7.

rosariopfernandes:
@samtstern that sounds right. Let's wait for an update from the crashlytics team

samtstern:
I think this is no longer needed:
https://github.com/firebase/quickstart-android/pull/1305

PR #1303, opened 2021-06-22
chore(*): disable jetifier

rosariopfernandes:
The main goal of this PR is to disable Jetifier in our project.
In order to make that happen, I also had to:
- upgrade facebook sdk to v8.1.0
- upgrade glide version (4.x)
- update workmanager to v2.5.0
- declare picasso 2.8 explicitly to be used in the FIAM quickstart


samtstern (approved):

samtstern:
@rosariopfernandes very nice! Impressed that this all builds.

rosariopfernandes:
I was hoping this would reduce build times, but apparently that's not the case :sweat_smile:

PR #1310, opened 2021-07-07
chore: remove NDK installation from GitHub Actions workflow

rosariopfernandes:
I believe the main reason why we had the Android NDK on our workflow was for the ML Kit samples to build.
Since these samples are now gone we no longer need the NDK on our CI.


samtstern:
Nice! I hated that step

samtstern (approved):

PR #1423, opened 2022-10-15
Menu option deprecation fix

nozotrox:
Hi @thatfiredev 
The deprecated methods have been replaced with the updated ones according to the solution  proposed. 

Thank you!


nozotrox:
Hi @thatfiredev 
I'll be closing this pull request because the CLA is failing as the account that is used in the commit is different from the one that i used to sign the CLA.

PR #1425, opened 2022-10-16
Menu option deprecation fix

nozotrox:
Hi @thatfiredev 
Sending the latest update with the commits with my current account.

Thanks (It was fun :) )


firestore/app/src/main/java/com/google/firebase/example/fireeats/java/MainFragment.java
```diff
@@ -169,7 +177,7 @@ public boolean onOptionsItemSelected(MenuItem item) {
                 startSignIn();
                 break;
         }
-        return super.onOptionsItemSelected(item);
+        return true;
```
thatfiredev, line 180: I have a question related to a code that was already in the repo and was unchanged by this PR, but since we're updating the menu, it might make sense to address it too.


Would it make sense to return instead of breaking here?
```java
        switch (item.getItemId()) {
            case R.id.menu_add_items:
                onAddItemsClicked();
                return true;
            case R.id.menu_sign_out:
                AuthUI.getInstance().signOut(requireContext());
                startSignIn();
                return true;
        }
        return false;
```

firestore/app/src/main/java/com/google/firebase/example/fireeats/kotlin/MainFragment.kt
```diff
@@ -123,20 +129,19 @@ class MainFragment : Fragment(),
...
+    override fun onMenuItemSelected(item: MenuItem): Boolean {
         when (item.itemId) {
             R.id.menu_add_items -> onAddItemsClicked()
             R.id.menu_sign_out -> {
                 AuthUI.getInstance().signOut(requireContext())
                 startSignIn()
             }
         }
-        return super.onOptionsItemSelected(item)
+        return true;
```
thatfiredev, line 144: Same suggestion (above) I wrote for Java applies here.

thatfiredev (changes requested):
@nozotrox Thanks for working on this. I left a few comments

database/app/src/main/java/com/google/firebase/quickstart/database/kotlin/MainFragment.kt
```diff
@@ -32,7 +34,10 @@ class MainFragment : Fragment() {
 
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
-        setHasOptionsMenu(true)
+
+        // MenuProvider
+        val menuHost: MenuHost = requireActivity() as MenuHost
+        menuHost.addMenuProvider(this);
```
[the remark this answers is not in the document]
nozotrox: Noted. I shall fix it.

firestore/app/src/main/java/com/google/firebase/example/fireeats/java/MainFragment.java
nozotrox: Yes, i makes more sense as it returns only true if the option exists or is intended to be selected. (That's my understanding)
I'll update as such.

Thanks.

firestore/app/src/main/java/com/google/firebase/example/fireeats/kotlin/MainFragment.kt
nozotrox: Noted.

database/gradle.properties
```diff
@@ -10,7 +10,7 @@
 # Specifies the JVM arguments used for the daemon process.
 # The setting is particularly useful for tweaking memory settings.
 org.gradle.jvmargs=-Xmx1536m
-
+android.useAndroidX=true
```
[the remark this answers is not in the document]
nozotrox: When trying to run the database app I was faced with the following error:
```
A problem occurred configuring project ':app'.
> org.gradle.api.GradleException: androidx.navigation.safeargs can only be used with an androidx project
 Try:
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.

* Exception is:
org.gradle.api.ProjectConfigurationException: A problem occurred configuring project ':app'.
	at org.gradle.configuration.project.LifecycleProjectEvaluator.wrapException(LifecycleProjectEvaluator.java:84)
	at org.gradle.configuration.project.LifecycleProjectEvaluator.addConfigurationFailure(LifecycleProjectEvaluator.java:77)
	at org.gradle.configuration.project.LifecycleProjectEvaluator.access$400(LifecycleProjectEvaluator.java:55)
	at org.gradle.configuration.project.LifecycleProjectEvaluator$NotifyAfterEvaluate.run(LifecycleProjectEvaluator.java:255)
	at org.gradle.internal.operations.DefaultBuildOperationRunner$1.execute(DefaultBuildOperationRunner.java:29)
	at org.gradle.internal.operations.DefaultBuildOperationRunner$1.execute(DefaultBuildOperationRunner.java:26)
...
```

To fix the error I used this proposed [solution](https://stackoverflow.com/a/55756802/10114458).
After adding the line under this comment the project was able to run.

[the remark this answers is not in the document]
thatfiredev: That's odd because it works fine for me. But I see no issues with including this flag. So let's proceed with it :)

firestore/app/src/main/java/com/google/firebase/example/fireeats/kotlin/MainFragment.kt
```diff
@@ -92,6 +94,10 @@ class MainFragment : Fragment(),
             }
         }
 
+        // MenuProvider
+        val menuHost : MenuHost = requireActivity() as MenuHost
```
thatfiredev, line 98: nit: the [Kotlin style guide](https://developer.android.com/kotlin/style-guide#horizontal) recommends only using a whitespace before a colon if in a class declaration.
```suggestion
        val menuHost: MenuHost = requireActivity() as MenuHost
```

thatfiredev (approved):
LGTM! Thanks for your contribution @nozotrox !

PR #1450, opened 2023-01-24
Compose workspace config

millandavid:
For review by Rosário Fernandes

Created an option to use Config using Jetpack Compose. 

Viewmodels is the next step to be implemented to this Quickstart rather than having the FirebaseActivity file. 

In the MainComposeActivity file that was created, please review Composables to advise on whether to continue passing Lambdas or if a Context would be better suitable in this scenario to pass.


config/app/build.gradle
```diff
@@ -29,6 +32,22 @@ android {
 
     buildFeatures {
         viewBinding = true
+        compose true
```
thatfiredev, line 35: nit: Let's use equals here to make it consistent with the line above it.
```suggestion
        compose = true
```

config/app/build.gradle
```diff
@@ -50,9 +70,20 @@ dependencies {
     // For an optimal experience using Remote Config, add the Firebase SDK
     // for Google Analytics. This is recommended, but not required.
     implementation 'com.google.firebase:firebase-analytics'
+    implementation 'androidx.compose.material3:material3:1.0.0-alpha02'
```
thatfiredev, line 73: We generally avoid using alpha versions on our quickstarts.
Can we use the stable version of this library instead?
```suggestion
    implementation 'androidx.compose.material3:material3:1.0.1'
```

config/app/src/main/AndroidManifest.xml
```diff
@@ -5,20 +5,26 @@
...
-
+        android:theme="@style/AppTheme">
+        <activity
+            android:name=".kotlin.MainComposeActivity"
+            android:exported="false"
+            android:label="@string/title_activity_main_compose"
+            android:theme="@style/AppTheme">
+            <meta-data
+                android:name="android.app.lib_name"
+                android:value="" />
```
thatfiredev, line 16: Why is this `meta-data` needed?

config/app/src/main/AndroidManifest.xml
```diff
@@ -5,20 +5,26 @@
         android:allowBackup="true"
         android:icon="@mipmap/ic_launcher"
         android:label="@string/app_name"
-        android:theme="@style/AppTheme" >
-
+        android:theme="@style/AppTheme">
+        <activity
+            android:name=".kotlin.MainComposeActivity"
+            android:exported="false"
+            android:label="@string/title_activity_main_compose"
```
thatfiredev, line 12: I believe we can use the `app_name` as title like we do in other activities.
```suggestion
            android:label="@string/app_name"
```

config/app/src/main/res/values/strings.xml
```diff
@@ -1,4 +1,5 @@
 <resources>
     <string name="app_name">Firebase Remote Config</string>
-    <string name="fetch_remote_welcome_message">fetch remote welcome</string>
+    <string name="fetch_remote_welcome_message">Fetch Remote Welcome</string>
+    <string name="title_activity_main_compose">MainComposeActivity</string>
```
thatfiredev, line 4: We can remove remove this string since we'll be using `app_name` as the screen title.

config/app/src/main/res/values/strings.xml
```diff
@@ -1,4 +1,5 @@
 <resources>
     <string name="app_name">Firebase Remote Config</string>
-    <string name="fetch_remote_welcome_message">fetch remote welcome</string>
+    <string name="fetch_remote_welcome_message">Fetch Remote Welcome</string>
```
thatfiredev, line 3:
```suggestion
    <string name="fetch_remote_welcome_message">Fetch remote welcome</string>
```

config/app/src/main/java/com/google/samples/quickstart/config/kotlin/MainComposeActivity.kt
```diff
@@ -0,0 +1,137 @@
...
+    Text(text = "Hello $name!")
+}
+
+@Preview(showBackground = true)
+@Composable
+fun DefaultPreview() {
+    ConfigTheme {
+        Greeting("Android")
+    }
+}
```
thatfiredev, line 137: Seems like we're not using the `Gretting` composable, I think it can be deleted.

config/app/src/main/java/com/google/samples/quickstart/config/kotlin/MainComposeActivity.kt
```diff
@@ -0,0 +1,137 @@
...
+        AppNameBanner()
+        Spacer(modifier = Modifier.height(24.dp))
+
+        Image(painter = painterResource(R.drawable.firebase_lockup_400), contentDescription = "")
+//        Spacer(modifier = Modifier.height(16.dp))
+
+        ConfigText(remoteConfigDisplayText = remoteConfigDisplayText)
+        Spacer(modifier = Modifier.height(160.dp))
+
+        ConfigButton(myClickEventMSGFetcher = buttonClickEventMSGFetcher)
```
thatfiredev, line 84: It looks like this is the only usage of `ConfigButton`. I would write the implementation here instead of delegating to separate function. WDYT?

thatfiredev: Same applies for `ConfigText`.

thatfiredev (commented):
@millandavid Thanks for your PR!

I've added some initial comments that I think we should address before going into the `ViewModel` step.

config/app/build.gradle
millandavid: Thanks! Totally missed this

config/app/build.gradle
millandavid: Yep! Stable version works. Will avoid alpha versions moving forward

config/app/src/main/AndroidManifest.xml
millandavid: Just tested it without and it worked

config/app/src/main/java/com/google/samples/quickstart/config/kotlin/MainComposeActivity.kt
millandavid: Hey Rosário, our thought process was that it keeps the structure of the display obvious and readable by putting the composable details in a separate function.

thatfiredev: I think that would make sense for Composables that group other composables, or composables that take specific types. 

For example, a `Progressbar` could come with a `Text`, so we need both composables grouped together and it would result in something like
```kotlin
CustomProgressBar(text: String, progress: Float) {
    Column {
        ProgressBar(progress)
        Text(text)
    }
}
````

Or a Card that takes a custom `Article` class as argument:
```kotlin
CustomCard(article: Article) {
    Column {
        Text(article.title)
        Text(article.body)
        Text(article.publishDate)
    }
}
```

But in this particular case, since it's just a simple composable (Button or Text), maybe having it inline with its parent will still be easy to read, no need to delegate to a separate function.

millandavid: Hi Rosário, thanks for the explanation. I think it does make sense in this case. Thanks for the feedback!

config/app/src/main/java/com/google/samples/quickstart/config/kotlin/MainComposeActivity.kt
```diff
@@ -0,0 +1,135 @@
...
+            text = remoteConfigDisplayText,
+            fontSize = 16.sp
+        )
+        Spacer(modifier = Modifier.height(160.dp))
+
+        // Button to fetch remote welcome
+        Button(
+            colors = ButtonDefaults.buttonColors(containerColor = colorResource(R.color.colorAccent)),
+            onClick = { remoteConfigViewModel.fetchConfig() }
+            // Calls function from MainComposeActivity to change display text
```
thatfiredev, line 111: I think ti
```suggestion
            onClick = {
                // Fetch config and update the display text
                remoteConfigViewModel.fetchConfig()
            }
```

millandavid: I like it, thanks! I will do the changes right now!

config/app/src/main/java/com/google/samples/quickstart/config/kotlin/MainComposeActivity.kt
```diff
@@ -0,0 +1,151 @@
...
+    val scaffoldState = rememberScaffoldState() // this contains the `SnackbarHostState`
+    val coroutineScope = rememberCoroutineScope()
+
+    Scaffold(
+        scaffoldState = scaffoldState,
+        topBar = {
+        TopAppBar(
+            backgroundColor = colorResource(R.color.colorPrimary)
+        ) {
+            androidx.compose.material.Text(
```
thatfiredev, line 95: We're already importing this Composable. Can it be just `Text`?
```suggestion
            Text(
```

thatfiredev (approved):
Thank you so much @millandavid !!

PR #1453, opened 2023-01-30
sample implementation of registration tokens for FCM

andreaowu:


messaging/app/build.gradle
```diff
@@ -71,6 +71,7 @@ dependencies {
     implementation 'com.google.firebase:firebase-installations-ktx:17.1.0'
 
     implementation 'androidx.work:work-runtime:2.7.1'
+    implementation 'com.google.firebase:firebase-firestore-ktx:24.1.0'
```
thatfiredev, line 74: nit: can we have this dependency together with the other Firebase dependencies (maybe a line before androidx.work or even before firebase-installations) and add a comment why it's needed (something along the lines of "used to store FCM registration tokens").

messaging/app/src/main/AndroidManifest.xml
```diff
@@ -36,8 +36,6 @@
         </activity>
 
         <activity android:name=".kotlin.MainActivity" />
-        <activity android:name=".java.MainActivity" />
```
thatfiredev, line 39: Was this change intentional?

messaging/app/src/main/AndroidManifest.xml
```diff
@@ -46,15 +44,6 @@
...
 
-        <!-- [START firebase_service] -->
-        <service
-            android:name=".java.MyFirebaseMessagingService"
-            android:exported="false">
-            <intent-filter>
-                <action android:name="com.google.firebase.MESSAGING_EVENT" />
-            </intent-filter>
-        </service>
-        <!-- [END firebase_service] -->
```
thatfiredev, line 57: Same as above: was this removal intentional?

messaging/app/src/main/java/com/google/firebase/quickstart/fcm/kotlin/MainActivity.kt
```diff
@@ -80,25 +88,46 @@ class MainActivity : AppCompatActivity() {
...
-            Firebase.messaging.getToken().addOnCompleteListener(OnCompleteListener { task ->
-                if (!task.isSuccessful) {
-                    Log.w(TAG, "Fetching FCM registration token failed", task.exception)
-                    return@OnCompleteListener
-                }
-
-                // Get new FCM registration token
-                val token = task.result
-
+            GlobalScope.launch {
```
thatfiredev, line 91: Note that the use of [GlobalScope is discouraged](https://developer.android.com/kotlin/coroutines/coroutines-best-practices#global-scope). Can we use [lifecycleScope](https://developer.android.com/topic/libraries/architecture/coroutines#lifecyclescope) instead? (you might need to add `androidx.lifecycle:lifecycle-runtime-ktx:2.4.0` as dependency to use that scope).

messaging/app/src/main/java/com/google/firebase/quickstart/fcm/kotlin/MainActivity.kt
```diff
@@ -11,11 +12,18 @@ import android.widget.Toast
...
+import com.google.firebase.firestore.ktx.firestore
 import com.google.firebase.ktx.Firebase
 import com.google.firebase.messaging.ktx.messaging
 import com.google.firebase.quickstart.fcm.R
 import com.google.firebase.quickstart.fcm.databinding.ActivityMainBinding
+import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.tasks.await
+import java.util.*
```
thatfiredev, line 26: I'm sure CI will complain about this, but unfortunately Android Studio does it automatically - wildcard imports are discouraged and result in build failures in our CI. Only import the classes and functions that you need (eg. `java.util.Calendar`) instead of the whole package (`java.util.*`).

messaging/app/src/main/java/com/google/firebase/quickstart/fcm/kotlin/MainActivity.kt
```diff
@@ -80,25 +88,46 @@ class MainActivity : AppCompatActivity() {
...
-            })
+                Log.d(TAG, "on click listener: $token")
+                val deviceToken = hashMapOf(
+                    "token" to token,
+                    "timestamp" to FieldValue.serverTimestamp(),
+                )
+
+                // Get user ID from Firebase Auth or your own server
+                Firebase.firestore.collection("fcmTokens").document("myuserid")
+                    .set(deviceToken)
```
thatfiredev, line 102: `.set()` runs asynchronously. And since we're already in a Coroutine scope, maybe we could await for its result too?
```suggestion
                Firebase.firestore.collection("fcmTokens").document("myuserid")
                    .set(deviceToken).await()
```

messaging/app/src/main/java/com/google/firebase/quickstart/fcm/kotlin/MainActivity.kt
```diff
@@ -80,25 +88,46 @@ class MainActivity : AppCompatActivity() {
...
         }
 
         Toast.makeText(this, "See README for setup instructions", Toast.LENGTH_SHORT).show()
         askNotificationPermission()
+
+        // In the app’s first Activity
+        val preferences = this.getPreferences(Context.MODE_PRIVATE)
+        val lastRefreshLong = preferences.getLong("lastRefreshDate", -1)
+        Log.d(TAG, "last refresh in preferences: $lastRefreshLong")
+        GlobalScope.launch {
```
thatfiredev, line 114: Same as above: use lifecycleScope instead.

messaging/app/src/main/java/com/google/firebase/quickstart/fcm/kotlin/MainActivity.kt
```diff
@@ -80,25 +88,46 @@ class MainActivity : AppCompatActivity() {
...
+            val document = Firebase.firestore.collection("refresh").document("refreshDate").get().await()
```
thatfiredev, line 115: Do we have a different collection (`refresh`) to refresh tokens? I assumed we only needed one collection (`fcmTokens`).

messaging/app/src/main/java/com/google/firebase/quickstart/fcm/kotlin/MainActivity.kt
```diff
@@ -80,25 +88,46 @@ class MainActivity : AppCompatActivity() {
...
+            val updatedTime = (document.data!!["lastRefreshDate"] as Timestamp).seconds * 1000
+            Log.d(TAG, "updatedTime: $updatedTime from Firestore")
+            val lastRefreshDate = Date(if (lastRefreshLong == -1L) updatedTime else lastRefreshLong)
+            val c = Calendar.getInstance()
+            c.time = lastRefreshDate
+            c.add(Calendar.DATE, 30)
```
thatfiredev, line 121: nit: In Kotlin you can use the `apply()` function to avoid repeating the `c` variable in every line:
```suggestion
            val c = Calendar.getInstance().apply {
                time = lastRefreshDate
                add(Calendar.DATE, 30)
            }
```

messaging/app/src/main/java/com/google/firebase/quickstart/fcm/kotlin/MainActivity.kt
```diff
@@ -115,6 +144,24 @@ class MainActivity : AppCompatActivity() {
         }
     }
 
+    private suspend fun getAndStoreRegToken(): String {
+        var token = ""
+        runBlocking {
```
thatfiredev, line 149: Why do we need `runBlocking` here? I believe it should work without it.

messaging/app/src/main/java/com/google/firebase/quickstart/fcm/kotlin/MyFirebaseMessagingService.kt
```diff
@@ -103,9 +106,18 @@ class MyFirebaseMessagingService : FirebaseMessagingService() {
      *
      * @param token The new token.
      */
-    private fun sendRegistrationToServer(token: String?) {
+    private fun storeToken(token: String?) {
```
thatfiredev, line 109: Can we keep the previous function name (`sendRegistrationToServer`) or use `sendTokenToServer`, I feel like it's more descriptive than `storeToken`.

messaging/app/src/main/java/com/google/firebase/quickstart/fcm/kotlin/MainActivity.kt
andreaowu: This is to store the last refresh date (since it's recommended that the token gets refreshed every month, I'm storing the last refresh date in Firestore), which is different than storing device tokens + refresh dates

messaging/app/src/main/java/com/google/firebase/quickstart/fcm/kotlin/MainActivity.kt
andreaowu: `Firebase.messaging.token.await()` needs `runBlocking`

thatfiredev: From what I see `getAndStoreRegToken()` is called from a Coroutine Scope, so I think we can make it a `suspend` function and remove `runBlocking`.

Calling `runBlocking` here would block the main thread until this function execution finished. Making it a suspend function would allow the function to execute on a coroutine instead. ([more about runBlocking](https://stackoverflow.com/a/52332499/5861618))

messaging/app/src/main/java/com/google/firebase/quickstart/fcm/kotlin/MainActivity.kt
```diff
@@ -115,6 +147,24 @@ class MainActivity : AppCompatActivity() {
...
+            Log.d(TAG, "getRegistrationToken(): $token")
+            // Add token and timestamp to Firestore for this user
+            val deviceToken = hashMapOf(
+                "token" to token,
+                "timestamp" to FieldValue.serverTimestamp(),
+            )
+
+            // Get user ID from Firebase Auth or your own server
+            Firebase.firestore.collection("fcmTokens").document("myuserid")
+                .set(deviceToken)
```
thatfiredev, line 163: Let's await for this too
```suggestion
                .set(deviceToken).await()
```

thatfiredev (commented):
@andreaowu CI is failing here due to trailing commas which was only introduced in Kotlin recently. We are running an old version of ktlint. I'll put up a PR to fix it

messaging/app/src/main/java/com/google/firebase/quickstart/fcm/kotlin/MainActivity.kt
thatfiredev: @andreaowu don't forget to replace this scope

thatfiredev: @andreaowu ping

thatfiredev (approved):
LGTM

Issue #1457, opened 2023-02-10
Create Jetpack Compose versions of the quickstarts


thatfiredev:
We're currently working on creating Jetpack Compose versions of each quickstarts. This public issue should serve as an Implementation Tracker for it.

We're moving all the calls to Firebase/Google Services from Kotlin Activities/Fragments to a ViewModel. The idea is that the current  Kotlin+XML quickstarts should share the same ViewModel as the Compose ones.

All the work is currently being done in the [compose branch](https://github.com/firebase/quickstart-android/tree/compose) and will be merged into the main branch once all quickstarts are done.

If you'd like to contribute to this effort, feel free to comment bellow which quickstart you'd like to work on and I'll assign it to you.

| Quickstart  | Assignee(s) | PR(s) | Status ( ⬜️  = not assigned, 🔄 = work in progress, ✅  = done) |
| ------------- | ------------- | ------------- | ------------- |
| admob  | @LokiFrostGiant  |  #1452   |  🔄   | 
| analytics  |   |   |  ⬜️  | 
| appdistribution  |   |    | ⬜️ | 
| auth  | @thatfiredev   |    |  🔄   | 
| config  | @millandavid  |  #1450  |  🔄   | 
| crash  |   |    |  ⬜️ | 
| database  |   |    |  ⬜️ | 
| dynamiclinks  |   |    |  ⬜️  | 
| firestore  |   |    |  ⬜️  | 
| functions  |   |    |  ⬜️  | 
| inappmessaging  |   |    |  ⬜️  | 
| messaging  |   |    |  ⬜️  | 
| perf  |   |    |  ⬜️  | 
| storage  |   |    |  ⬜️  |

millandavid:
I would like to work on crash please!

LokiFrostGiant:
I would like to work on storage, please 😎

thatfiredev:
@millandavid @LokiFrostGiant Assigned both of you. Go for it :)

Joel-K-Muraguri:
I would like to work on firestore, please.

thatfiredev:
@Joel-K-Muraguri Sure, assigned to you. Be sure to checkout the  [compose branch](https://github.com/firebase/quickstart-android/tree/compose)

Eff4real:
I would like to work on storage.

thatfiredev:
@Eff4real I'm afraid Storage is already taken. You may choose a quickstart that doesn't have any assignee.

[...]

PR #1459, opened 2023-02-16
[WIP] move firebase auth calls to ViewModels

thatfiredev:
Part of #1457 

---

Some caveats found in the process:

#### Google Sign in
Ideally, we'd inject the `GoogleSignInClient` the same way we inject `FirebaseAuth` into the ViewModel, but some APIs require an Activity, so I've decided to keep `GoogleSignInClient` on the UI code for now.


wallyf1972 (commented):
Leave my stuff alone

PR #1463, opened 2023-03-06
feat(appdistribution): add in-app-feedback support

thatfiredev:


appdistribution/app/src/main/java/com/google/firebase/appdistributionquickstart/java/MainActivity.java
```diff
@@ -21,6 +41,19 @@ protected void onCreate(Bundle savedInstanceState) {
         setContentView(binding.getRoot());
 
         mFirebaseAppDistribution = FirebaseAppDistribution.getInstance();
+
+        binding.btShowNotification.setOnClickListener(view -> {
+            mFirebaseAppDistribution.showFeedbackNotification(
+                    "Data Collection Notice",
```
lfkellogg, line 47: Could you use the same example text as we use here: https://github.com/firebase/firebase-android-sdk/blob/46e7a69781b6651582be2ea9413cd6c30913b829/firebase-appdistribution/test-app/src/main/res/values/strings.xml#L3

If you'd rather use a string literal instead of a resource, feel free to leave out the "link" part.

appdistribution/app/src/main/java/com/google/firebase/appdistributionquickstart/java/MainActivity.java
```diff
@@ -21,6 +41,19 @@ protected void onCreate(Bundle savedInstanceState) {
...
+                    InterruptionLevel.HIGH
+            );
+        });
+
+        binding.btSendFeedback.setOnClickListener(view -> {
+            mFirebaseAppDistribution.startFeedback("Thanks for sharing your feedback with us");
```
lfkellogg, line 53: Same comment as above about the example text.

appdistribution/app/src/main/java/com/google/firebase/appdistributionquickstart/java/MainActivity.java
```diff
@@ -37,4 +70,29 @@ public void onResume() {
...
+        // This is only necessary for API level >= 33 (TIRAMISU)
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
+            if (ContextCompat.checkSelfPermission(this, Manifest.permission.POST_NOTIFICATIONS) ==
+                    PackageManager.PERMISSION_GRANTED) {
+                // All set. We can post notifications
+            } else if (shouldShowRequestPermissionRationale(Manifest.permission.POST_NOTIFICATIONS)) {
+                // Display an educational UI explaining to the user the features that will be enabled
+                // by them granting the POST_NOTIFICATION permission. This UI should provide the user
+                // "OK" and "No thanks" buttons. If the user selects "OK," directly request the permission.
+                // If the user selects "No thanks," allow the user to continue without notifications.
```
lfkellogg, line 91: See here for an example from the test app in the SDK repo: https://github.com/firebase/firebase-android-sdk/blob/46e7a69781b6651582be2ea9413cd6c30913b829/firebase-appdistribution/test-app/src/main/kotlin/com/googletest/firebase/appdistribution/testapp/MainActivity.kt#L149

appdistribution/app/src/main/java/com/google/firebase/appdistributionquickstart/java/MainActivity.java
```diff
@@ -21,6 +41,19 @@ protected void onCreate(Bundle savedInstanceState) {
...
+                    "Data Collection Notice",
+                    InterruptionLevel.HIGH
+            );
+        });
+
+        binding.btSendFeedback.setOnClickListener(view -> {
+            mFirebaseAppDistribution.startFeedback("Thanks for sharing your feedback with us");
+        });
+
+        askNotificationPermission();
```
lfkellogg, line 56: I think a better modes for requesting permissions would be to call `askNotificationPermission()` only when the tester clicks the `btShowNotification` button. That would be more in line with the runtime permissions guidance: https://developer.android.com/training/permissions/requesting#workflow_for_requesting_permissions

In that case, you would call `FirebaseAppDistribution.showFeedbackNotification()` in the "// All set. We can post notifications" case in `askNotificationPermission()` below, and in the `sGranted == true` case in the `requestPermissionLauncher` initializer above.

lfkellogg (commented):
Nice! Left some comments.

By the way, do you intend to update the README as well? https://github.com/firebase/quickstart-android/blob/master/appdistribution/README.md

appdistribution/app/src/main/java/com/google/firebase/appdistributionquickstart/java/MainActivity.java
thatfiredev: Added as a resource.

appdistribution/app/src/main/java/com/google/firebase/appdistributionquickstart/java/MainActivity.java
thatfiredev: Thanks! I added that same dialog.

appdistribution/app/src/main/java/com/google/firebase/appdistributionquickstart/java/MainActivity.java
thatfiredev: Good one! I added a `showFeedbackNotification()` function to call it in both cases.

thatfiredev:
@lfkellogg Great callout on the README. I updated it and its screenshot. PTAL

appdistribution/app/src/main/java/com/google/firebase/appdistributionquickstart/java/MainActivity.java
```diff
@@ -37,4 +70,44 @@ public void onResume() {
                     }
                 });
     }
+
+    @Override
+    protected void onDestroy() {
+        super.onDestroy();
+        // Hide the notification once this activity is destroyed
+        mFirebaseAppDistribution.cancelFeedbackNotification();
+    }
```
lfkellogg, line 79: You shouldn't need this anymore now that we made the change to automatically hide the notification

appdistribution/app/src/main/java/com/google/firebase/appdistributionquickstart/java/MainActivity.java
```diff
@@ -37,4 +70,44 @@ public void onResume() {
...
+    private void showFeedbackNotification() {
+        mFirebaseAppDistribution.showFeedbackNotification(
+                R.string.feedbackAdditionalFormText,
+                InterruptionLevel.HIGH
+        );
+    }
+
+    private void askNotificationPermission() {
+        // This is only necessary for API level >= 33 (TIRAMISU)
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
```
lfkellogg, line 90: I don't think you want this check here anymore. Even if they are on an older device we still want to show the notification. On older devices the `checkSelfPermission()` call will return `true`.

appdistribution/app/src/main/res/values/strings.xml
```diff
@@ -1,6 +1,9 @@
 <resources>
     <string name="app_name">App Distribution Quickstart</string>
-    <string name="textview_text">Welcome to the Firebase App Distribution Quickstart app. Press the button to trigger an analytics event!</string>
+    <string name="textview_text">Welcome to the Firebase App Distribution Quickstart app</string>
 
     <string name="installation_id_fmt">Firebase Installation ID: %s</string>
+    <string name="show_feedback_notification">Show Feedback Notification</string>
+    <string name="send_feedback">Send Feedback</string>
+    <string name="feedbackAdditionalFormText">This is a placeholder for text from the developer that is shown to the tester before they provide feedback. It can contain <a href="https://firebase.google.com/docs/app-distribution">links</a> to more information.</string>
```
lfkellogg, line 8: The rest of the IDs are `snake_case`. Might as well be consistent.

appdistribution/app/src/main/java/com/google/firebase/appdistributionquickstart/kotlin/KotlinMainActivity.kt
```diff
@@ -34,10 +67,49 @@ class KotlinMainActivity : AppCompatActivity() {
             }
     }
 
+    override fun onDestroy() {
+        super.onDestroy()
+        // Hide the notification once this activity is destroyed
+        firebaseAppDistribution.cancelFeedbackNotification()
```
marinacoelho, line 73: According to Lee's comment above, I think we don't need this line here as well.

appdistribution/app/src/main/java/com/google/firebase/appdistributionquickstart/kotlin/KotlinMainActivity.kt
```diff
@@ -34,10 +67,49 @@ class KotlinMainActivity : AppCompatActivity() {
...
+    private fun showFeedbackNotification() {
+        firebaseAppDistribution.showFeedbackNotification(
+            R.string.feedbackAdditionalFormText,
+            InterruptionLevel.HIGH
+        )
+    }
+
+    private fun askNotificationPermission() {
+        // This is only necessary for API level >= 33 (TIRAMISU)
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
```
marinacoelho, line 85: Same as Lee's comment above. And if we absolutely need to keep this check here, what do you think of changing this function a bit to use `when` instead of `if` / `else if` / `else` ? (Or `switch` / `case` on the Java file)/

appdistribution/app/src/main/java/com/google/firebase/appdistributionquickstart/java/MainActivity.java
thatfiredev: When does the automatic hide take place? I tried closing the Activity, but it didn't dismiss the notification. Is it when the app is closed?

appdistribution/app/src/main/res/values/strings.xml
thatfiredev: Done

appdistribution/app/src/main/java/com/google/firebase/appdistributionquickstart/java/MainActivity.java
thatfiredev: cc: @marinacoelho 
The check is due to the `POST_NOTIFICATIONS` field which was introduced in API 33.
I have added an else branch to display the feedback notification on older devices.

Screenshot 2023-03-07 at 13 12 01

appdistribution/app/src/main/java/com/google/firebase/appdistributionquickstart/kotlin/KotlinMainActivity.kt
thatfiredev: Having a `when` here might work for Kotlin, since the language allows usage of `when` without a subject. But not sure if Java has a switch-case without a subject 🤔

PR #1477, opened 2023-04-13
Compose Firebase Dynamic Links

argzdev:
For review by Rosário Fernandes

- Add Jetpack Compose option in `EntryChoiceActivity` 
- Removed business logic from `MainActivity`
- Added `DynamicLinksViewModel` for business logic
- Added `MainComposeActivity` for the Jetpack Compose UI

### DIFF
--- a/dynamiclinks/app/src/main/java/com/google/firebase/quickstart/deeplinks/kotlin/DynamicLinksViewModel.kt
+++ b/dynamiclinks/app/src/main/java/com/google/firebase/quickstart/deeplinks/kotlin/DynamicLinksViewModel.kt
@@ -6,6 +6,7 @@ import android.util.Log
 import androidx.annotation.VisibleForTesting
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.ViewModelProvider
+import androidx.lifecycle.viewModelScope
 import androidx.lifecycle.viewmodel.CreationExtras
 import com.google.firebase.dynamiclinks.FirebaseDynamicLinks
 import com.google.firebase.dynamiclinks.PendingDynamicLinkData
@@ -13,6 +14,8 @@ import com.google.firebase.dynamiclinks.ktx.*
 import com.google.firebase.ktx.Firebase
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.tasks.await
 
 class DynamicLinksViewModel(
     private val dynamicLinks: FirebaseDynamicLinks
@@ -28,10 +31,12 @@ class DynamicLinksViewModel(
     val validUriPrefix: StateFlow<Boolean> = _validUriPrefix
 
     fun getDynamicLink(intent: Intent) {
-        dynamicLinks
-            .getDynamicLink(intent)
-            .addOnSuccessListener { pendingDynamicLinkData: PendingDynamicLinkData? ->
-                // Get deep link from result (may be null if no link is found)
+        viewModelScope.launch {
+            try {
+                val pendingDynamicLinkData: PendingDynamicLinkData = dynamicLinks
+                    .getDynamicLink(intent)
+                    .await()
+
                 var deepLink: Uri? = null
                 if (pendingDynamicLinkData != null) {
                     deepLink = pendingDynamicLinkData.link
@@ -48,8 +53,10 @@ class DynamicLinksViewModel(
                 } else {
                     Log.d(TAG, "getDynamicLink: no link found")
                 }
+            } catch (e: Exception) {
+                Log.w(TAG, "getDynamicLink:onFailure", e)
             }
-            .addOnFailureListener { e: Exception -> Log.w(TAG, "getDynamicLink:onFailure", e) }
+        }
     }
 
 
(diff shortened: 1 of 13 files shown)


thatfiredev:
@argzdev I'll take a closer look at these, but in the meantime I noticed that CI is failing due to snippets errors (the `[START] and `[END]` comments). Feel free to remove those, we're no longer using them.

argzdev:
Got it, I'll remove those snippets. Thanks for taking a look into it, @thatfiredev!

argzdev:
In the `MainComposeActivity`, should we add a `@Preview`? I think we can just separate the content from the `MainAppView`, so we can add a `@Preview` of the layout.

dynamiclinks/build.gradle
```diff
@@ -1,6 +1,9 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 
 buildscript {
+    ext {
+        compose_version = '1.3.0'
+    }
```
thatfiredev, line 6: Do we really need this?

I assumed the module would pick up the `compose_version` declared in the root-level build.gradle file:

https://github.com/firebase/quickstart-android/blob/b634e6269981ea3767db51e6a1a3f2ed79706548/build.gradle#L3

dynamiclinks/app/src/main/java/com/google/firebase/quickstart/deeplinks/kotlin/DynamicLinksViewModel.kt
```diff
@@ -0,0 +1,129 @@
...
+    val shortLink: StateFlow<String> = _shortLink
+
+    private val _validUriPrefix = MutableStateFlow(true)
+    val validUriPrefix: StateFlow<Boolean> = _validUriPrefix
+
+    fun getDynamicLink(intent: Intent) {
+        dynamicLinks
+            .getDynamicLink(intent)
+            .addOnSuccessListener { pendingDynamicLinkData: PendingDynamicLinkData? ->
+                // Get deep link from result (may be null if no link is found)
```
thatfiredev, line 34: Can we use Coroutines instead of the completion listeners?

That would be something like:

```kotlin
fun getDynamicLink(intent: Intent) {
    viewModelScope.launch {
        try {
            val pendingDynamicLinkData: PendingDynamicLinkData? = dynamicLinks
                .getDynamicLink(intent)
                .await()
            // Get deep link from result (may be null if no link is found)
            var deepLink: Uri? = null
            if (pendingDynamicLinkData != null) {
                deepLink = pendingDynamicLinkData.link
            }

            ... (the rest of addOnSuccessListener)
        } catch (e) {
            Log.w(TAG, "getDynamicLink failure", e)
        }
    }
}
```


The reason for that is that we're trying to promote Coroutines as a best practice.

dynamiclinks/app/src/main/java/com/google/firebase/quickstart/deeplinks/kotlin/DynamicLinksViewModel.kt
```diff
@@ -0,0 +1,129 @@
...
+            link = deepLink
+            domainUriPrefix = uriPrefix
+            androidParameters {
+                minimumVersion = minVersion
+            }
+        }.addOnSuccessListener { (shortLink, flowchartLink) ->
+            _shortLink.value = shortLink.toString()
+        }.addOnFailureListener { e ->
+            Log.e(TAG, e.toString())
+        }
```
thatfiredev, line 103: Can we use Coroutines instead?

dynamiclinks/app/src/main/java/com/google/firebase/quickstart/deeplinks/kotlin/MainComposeActivity.kt
```diff
@@ -0,0 +1,293 @@
...
+            }
+        }
+    }
+}
+
+@Composable
+fun MainAppView(
+    lifecycleOwner: LifecycleOwner = LocalLifecycleOwner.current,
+    dynamicLinksViewModel: DynamicLinksViewModel = viewModel(factory = DynamicLinksViewModel.Factory)
+){
```
thatfiredev, line 62:
```suggestion
) {
```

dynamiclinks/app/src/main/java/com/google/firebase/quickstart/deeplinks/kotlin/MainComposeActivity.kt
```diff
@@ -0,0 +1,293 @@
...
+                Text(
+                    text = stringResource(R.string.app_name),
+                    style = MaterialTheme.typography.h6,
+                    textAlign = TextAlign.Center,
+                    modifier = Modifier.padding(8.dp),
+                    color = Color.White
+                )
+            }
+        },
+        content = { it
```
thatfiredev, line 128: This `it` seems like it's not being used here. Is it needed?
```suggestion
        content = {
```

argzdev: Yeah what I did here was just consume it without using it, because since Compose 1.2.0 it's required to use padding parameter to be passed into Scaffold content composable. In my next commit, what I did instead is to use a column to be able to use the paddingValue from `it`.

dynamiclinks/app/src/main/java/com/google/firebase/quickstart/deeplinks/kotlin/DynamicLinksViewModel.kt
argzdev: Okay, got it

dynamiclinks/app/src/main/java/com/google/firebase/quickstart/deeplinks/kotlin/DynamicLinksViewModel.kt
argzdev: Got it, will do.

dynamiclinks/app/src/main/java/com/google/firebase/quickstart/deeplinks/kotlin/MainComposeActivity.kt
argzdev: Good catch! I didn't even notice the space.

thatfiredev (approved):
LGTM

Thanks @argzdev !

PR #1483, opened 2023-04-24
Remove unnecessary import

Leesin0222:
Unnecessary import has been removed


thatfiredev (approved):
Good catch, thanks @Leesin0222 !

PR #1484, opened 2023-04-25
unified imports

Leesin0222:
Currently, if you look at all EntryChoiceActivity, most of them use unifying MainActivity.

And in some apps, you can also see that it's divided into MainActivity and Kotlin MainActivity
I thought if it wasn't named like the picture below, I might be confused if it was java or kotlin or both Kotlin MainActivity.
image

<br/>
So I unified the import so that developers could easily tell if it was java's MainActivity or Kotlin's MainActivity.


thatfiredev (approved):
@Leesin0222 We plan to delete the java examples at some point so this will no longer be an issue. But I will accept this PR for now. Thanks!

PR #1516, opened 2023-07-24
Use key constant when checking if welcome_message was updated

dazza5000:


thatfiredev (approved):
Thank you @dazza5000 !

PR #1606, opened 2024-04-24
Configuration of firebase.json due to firebase tools 13.0.0 breaking change

lehcar09:
In relation to firebase-tools 13.0.0 breaking change https://github.com/firebase/firebase-tools/pull/6555, Firebase CLI will only reserve functions directory when there is a functions configuration entry in firebase.json that does not specify the source directory.

### DIFF
--- a/functions/firebase.json
+++ b/functions/firebase.json
@@ -1,1 +1,14 @@
-{}
+{
+  "functions": [
+    {
+      "source": "functions",
+      "codebase": "default",
+      "ignore": [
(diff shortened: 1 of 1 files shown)


thatfiredev (approved):
Thanks @lehcar09 !

Can you please check the failing CLA? I'm not sure why it's failing, as it should always succeed for Googlers 🤔

thatfiredev (approved):
LGTM. Thanks @lehcar09 !

PR #1611, opened 2024-05-10
Update model to 1.5-pro

rlazo:


vertexai/app/src/main/kotlin/com/google/firebase/quickstart/vertexai/GenerativeAiViewModelFactory.kt
```diff
@@ -46,17 +46,17 @@ val GenerativeViewModelFactory = object : ViewModelProvider.Factory {
...
 
                 isAssignableFrom(PhotoReasoningViewModel::class.java) -> {
-                    // Initialize a GenerativeModel with the `gemini-pro-vision` AI model
+                    // Initialize a GenerativeModel with the `gemini-pro` AI model
                     // for multimodal text generation
                     val generativeModel = Firebase.vertexAI.generativeModel(
-                        modelName = "gemini-1.0-pro-vision",
+                        modelName = "gemini-1.5-pro-preview-0409",
                         generationConfig = config
                     )
```
thatfiredev, line 61: At this point I think we can initialize the model once (like before the return clause) and then pass to every ViewModel. The reason why I had different initializations was because of `gemini-pro-vision`. But now that we use the same model for every use case, maybe we could simplify our code?

Stromlight (commented):
I think anyway the model going to be same so their can focus on which is appropriate for use is necessary to add.

thatfiredev (approved):
I'll merge as is for now. We can discuss code reuse in a separate space. :)

PR #1612, opened 2024-05-10
Remove properties not supported by java 17

rlazo:


vertexai/gradle.properties
```diff
@@ -2,9 +2,7 @@
 # http://www.gradle.org/docs/current/userguide/build_environment.html
 #
 # Specifies the JVM arguments used for the daemon process.
-# The setting is particularly useful for tweaking memory settings.
-# Default value: -Xmx1024m -XX:MaxPermSize=256m
- org.gradle.jvmargs=-Xmx2048m -XX:MaxPermSize=512m -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8
+org.gradle.jvmargs=-Xmx8g
```
thatfiredev, line 5: It seems like 2048m is the default set by Android Studio and it was the default that the sample was using. Should we stick with that?

```suggestion
org.gradle.jvmargs=-Xmx2048m
```

thatfiredev (approved):

MWGW (approved):

PR #1653, opened 2024-08-12
VertexAI: add audio input feature

tanzimfh:
This adds a new feature to the Vertex AI sample app that lets the user record audio and ask the model a question about it.

### DIFF
--- a/vertexai/app/build.gradle.kts
+++ b/vertexai/app/build.gradle.kts
@@ -62,6 +62,7 @@ dependencies {
     implementation("androidx.compose.ui:ui-graphics")
     implementation("androidx.compose.ui:ui-tooling-preview")
     implementation("androidx.compose.material3:material3")
+    implementation("androidx.compose.material:material-icons-extended")
     implementation("io.coil-kt:coil-compose:2.7.0")
     
     implementation("com.google.firebase:firebase-analytics:22.0.2")
--- a/vertexai/app/build.gradle.kts
+++ b/vertexai/app/build.gradle.kts
@@ -62,7 +62,6 @@ dependencies {
     implementation("androidx.compose.ui:ui-graphics")
     implementation("androidx.compose.ui:ui-tooling-preview")
     implementation("androidx.compose.material3:material3")
-    implementation("androidx.compose.material:material-icons-extended")
     implementation("io.coil-kt:coil-compose:2.7.0")
     
     implementation("com.google.firebase:firebase-analytics:22.0.2")
--- a/vertexai/app/src/main/kotlin/com/google/firebase/quickstart/vertexai/GenerativeAiViewModelFactory.kt
+++ b/vertexai/app/src/main/kotlin/com/google/firebase/quickstart/vertexai/GenerativeAiViewModelFactory.kt
@@ -20,6 +20,7 @@ import androidx.lifecycle.ViewModel
 import androidx.lifecycle.ViewModelProvider
 import androidx.lifecycle.viewmodel.CreationExtras
 import com.google.firebase.Firebase
+import com.google.firebase.quickstart.vertexai.feature.audio.AudioViewModel
 import com.google.firebase.quickstart.vertexai.feature.chat.ChatViewModel
 import com.google.firebase.quickstart.vertexai.feature.functioncalling.FunctionsChatViewModel
 import com.google.firebase.quickstart.vertexai.feature.multimodal.PhotoReasoningViewModel
@@ -94,6 +95,15 @@ val GenerativeViewModelFactory = object : ViewModelProvider.Factory {
                         tools = tools
                     )
                     FunctionsChatViewModel(generativeModel)
+                }
+
+                isAssignableFrom(AudioViewModel::class.java) -> {
+                    // Initialize a GenerativeModel with the `gemini-pro` AI model for audio generation
+                    val generativeModel = Firebase.vertexAI.generativeModel(
+                        modelName = "gemini-1.5-pro-001",
+                        generationConfig = config
+                    )
+                    AudioViewModel(generativeModel)
                 }
 
                 else ->
--- a/vertexai/app/src/main/kotlin/com/google/firebase/quickstart/vertexai/feature/audio/AudioRecorder.kt
+++ b/vertexai/app/src/main/kotlin/com/google/firebase/quickstart/vertexai/feature/audio/AudioRecorder.kt
@@ -22,7 +22,6 @@ import android.os.Build
 import java.io.File
 
 class AudioRecorder(private val context: Context) {
-    val mimeType = "audio/aac"
     private var recorder: MediaRecorder? = null
     private var outputFilePath: String? = null
 
--- a/vertexai/app/src/main/kotlin/com/google/firebase/quickstart/vertexai/feature/audio/AudioRecorder.kt
+++ b/vertexai/app/src/main/kotlin/com/google/firebase/quickstart/vertexai/feature/audio/AudioRecorder.kt
@@ -19,7 +19,6 @@ package com.google.firebase.quickstart.vertexai.feature.audio
 import android.content.Context
 import android.media.MediaRecorder
 import android.os.Build
-import android.os.Environment
 import java.io.File
 
 class AudioRecorder(private val context: Context) {
@@ -28,9 +27,9 @@ class AudioRecorder(private val context: Context) {
     private var outputFilePath: String? = null
 
     fun startRecording() {
-        val audioFileName = "recording_${System.currentTimeMillis()}.m4a"
-        val musicDir = context.getExternalFilesDir(Environment.DIRECTORY_MUSIC)
-        outputFilePath = File(musicDir, audioFileName).absolutePath
+        outputFilePath = File.createTempFile(
+            "recording_${System.currentTimeMillis()}", ".m4a", context.cacheDir
+        ).absolutePath
 
         recorder = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S) {
             MediaRecorder(context)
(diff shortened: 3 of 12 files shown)


vertexai/app/src/main/kotlin/com/google/firebase/quickstart/vertexai/feature/audio/AudioViewModel.kt
```diff
@@ -0,0 +1,62 @@
...
+    val uiState: StateFlow<AudioUiState> = _uiState.asStateFlow()
+
+    fun reason(
+        userInput: String,
+        mimeType: String,
+        audioBytes: ByteArray,
+    ) {
+        _uiState.value = AudioUiState.Loading
+        val prompt = if (userInput.isBlank()) "Answer the question in the audio."
+        else "Listen to the audio, and then answer the following question: $userInput"
```
thatfiredev, line 43: We might want to have braces here, for better readability.
From the [Kotlin style guide](https://developer.android.com/kotlin/style-guide#braces):

> Braces are not required for `when` branches and `if` expressions which have no more than one `else` branch and which fit on a single line.
Braces are otherwise required for any `if`, `for`, `when` branch, `do`, and `while` statements and expressions, even when the body is empty or contains only a single statement.

```suggestion
        val prompt = if (userInput.isBlank()) {
          "Answer the question in the audio."
        } else {
          "Listen to the audio, and then answer the following question: $userInput"
        }
```

vertexai/app/build.gradle.kts
```diff
@@ -62,6 +62,7 @@ dependencies {
     implementation("androidx.compose.ui:ui-graphics")
     implementation("androidx.compose.ui:ui-tooling-preview")
     implementation("androidx.compose.material3:material3")
+    implementation("androidx.compose.material:material-icons-extended")
```
thatfiredev, line 65: Quick note(no action needed): usage of this package is [discouraged](https://mvnrepository.com/artifact/androidx.compose.material/material-icons-extended) due to its large size:

> This module contains all Material icons. It is a very large dependency and should not be included directly.

But I noticed that the Android team uses in their own [samples](https://github.com/search?q=repo%3Aandroid%2Fcompose-samples%20material-icons-extended&type=code), so I think it should be fine if we use for this one.

vertexai/app/src/main/kotlin/com/google/firebase/quickstart/vertexai/feature/audio/AudioRecorder.kt
```diff
@@ -0,0 +1,68 @@
...
+package com.google.firebase.quickstart.vertexai.feature.audio
+
+import android.content.ContentValues
+import android.content.Context
+import android.media.MediaRecorder
+import android.os.Environment
+import android.provider.MediaStore
+
+class AudioRecorder(private val context: Context) {
+    val mimeType = "audio/aac"
```
thatfiredev, line 26: It seems like we never actually change the mimeType? Should we remove this variable from the `AudioRecorder` class and just pass it to `content { }` in the ViewModel ?

rlazo, line 26: we have audio/aac here, and audio/mp4 in line 35. Should they match?

vertexai/app/src/main/kotlin/com/google/firebase/quickstart/vertexai/feature/audio/AudioRecorder.kt
```diff
@@ -0,0 +1,68 @@
...
+    private var recorder: MediaRecorder? = null
+    private var outputFilePath: String? = null
+
+    fun startRecording() {
+        val audioFileName = "recording_${System.currentTimeMillis()}.m4a"
```
rlazo, line 31: Does this mean that the file gets stored forever on the device?

rlazo:
If you click record, rotate the device, and then click stop, the app crashes.

vertexai/app/build.gradle.kts
```diff
@@ -27,7 +27,7 @@ android {
 
     defaultConfig {
         applicationId = "com.google.firebase.quickstart.vertexai"
-        minSdk = 26
+        minSdk = 31
```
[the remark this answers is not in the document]
tanzimfh: I was trying to avoid using the deprecated [`MediaRecorder()` constructor](https://developer.android.com/reference/kotlin/android/media/MediaRecorder#mediarecorder_1), but I have reverted this bump and added a condition to use the deprecated constructor on sdk < 31.

vertexai/app/src/main/kotlin/com/google/firebase/quickstart/vertexai/GenerativeAiViewModelFactory.kt
```diff
@@ -96,6 +97,15 @@ val GenerativeViewModelFactory = object : ViewModelProvider.Factory {
                     FunctionsChatViewModel(generativeModel)
                 }
 
+                isAssignableFrom(AudioViewModel::class.java) -> {
+                    // Initialize a GenerativeModel with the `gemini-pro` AI model for audio generation
+                    val generativeModel = Firebase.vertexAI.generativeModel(
+                        modelName = "gemini-1.5-pro-001",
```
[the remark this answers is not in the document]
tanzimfh: That's a very good question.

According to the [docs](https://cloud.google.com/vertex-ai/generative-ai/docs/multimodal/audio-understanding#audio_requirements) Flash can also process `audio/aac`, but I get a response like this:
image

vertexai/app/build.gradle.kts
tanzimfh: I only need it for the `Mic` and `Stop` (which is just a square) icons. Is making a local copy a better option?

vertexai/app/src/main/kotlin/com/google/firebase/quickstart/vertexai/feature/audio/AudioRecorder.kt
tanzimfh: I added it here to keep all mentions of the encoding format in the same file for lower coupling, but I can hard-code it into the ViewModel if you prefer.

tanzimfh: My understanding was that the mime type on line 35 should reflect the container format based on [this](https://developer.android.com/reference/android/provider/MediaStore.MediaColumns.html#MIME_TYPE):
> This is typically defined based on the file extension of the media item.

but I'm removing the use of MediaStore and line 35 along with it

vertexai/app/src/main/kotlin/com/google/firebase/quickstart/vertexai/feature/audio/AudioRecorder.kt
tanzimfh: The file gets deleted on [line 65](https://github.com/firebase/quickstart-android/pull/1653/files/7a046b4db5d693e6269d47e97ef012a693a1682b#diff-9cf78f53c605c40639d5acc59bd725eb868e4aa130a90f8bceed912ecba1846dR65) after being read.

I was running into issues trying to avoid using file storage, but I can continue looking into it if you think it's worthwhile.

rlazo: If the app crashes for any reason before getting to that point, the file will remain there. 

You would ideally use a temp file in the cache 

https://stackoverflow.com/questions/3425906/creating-temporary-files-in-android

But you can add a TODO to work on this later

vertexai/app/src/main/kotlin/com/google/firebase/quickstart/vertexai/feature/audio/AudioRecorder.kt
rlazo: sgtm

vertexai/app/src/main/kotlin/com/google/firebase/quickstart/vertexai/feature/audio/AudioRecorder.kt
tanzimfh: updated, PTAL: https://github.com/firebase/quickstart-android/pull/1653/commits/36989b5b91f59afbc8fe1fe7bb5d334aa4472550

vertexai/app/build.gradle.kts
thatfiredev: Yes, making a copy is usually better.

vertexai/app/src/main/kotlin/com/google/firebase/quickstart/vertexai/GenerativeAiViewModelFactory.kt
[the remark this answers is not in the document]
thatfiredev: @tanzimfh Oof... Let's go with Pro then.

I'll ask the team if we should update those docs

vertexai/app/src/main/kotlin/com/google/firebase/quickstart/vertexai/feature/audio/AudioRecorder.kt
thatfiredev: @tanzimfh I think that might be better and easier to find for people looking for an example.

vertexai/app/src/main/kotlin/com/google/firebase/quickstart/vertexai/feature/audio/AudioScreen.kt
```diff
@@ -0,0 +1,239 @@
...
+                            } else audioData = null
+                        },
+                        modifier = Modifier
+                            .padding(all = 4.dp)
+                            .align(Alignment.CenterVertically),
+                    ) {
+                        Icon(
+                            imageVector = if (isRecording) ImageVector.vectorResource(R.drawable.stop)
+                            else if (audioData == null) ImageVector.vectorResource(R.drawable.mic)
+                            else Icons.Filled.Delete,
```
thatfiredev, line 136:
```suggestion
                            imageVector = if (isRecording) {
                                ImageVector.vectorResource(R.drawable.stop)
                            } else if (audioData == null) {
                                ImageVector.vectorResource(R.drawable.mic)
                            } else {
                                Icons.Filled.Delete
                            },
```

thatfiredev (approved):
LGTM

PR #1654, opened 2024-08-20
Compose code for analytics

lehcar09:
Initial commit with fixes from previous review

Initial changes made:
Add Jetpack Compose option in EntryChoiceActivity
Removed business logic from MainActivity
Added FirebaseAnalyticsViewModel for business logic
Added ComposeMainActivity for the Jetpack Compose UI

Changes made after review
-Rename recordShare API to recordShareEvent
-Pass screenClass parameter in recordScreenView

### DIFF
--- a/analytics/app/src/main/java/com/google/firebase/quickstart/analytics/kotlin/FirebaseAnalyticsViewModel.kt
+++ b/analytics/app/src/main/java/com/google/firebase/quickstart/analytics/kotlin/FirebaseAnalyticsViewModel.kt
@@ -1,0 +1,83 @@
+package com.google.firebase.quickstart.analytics.kotlin
+
+import android.content.Context
+import androidx.compose.runtime.MutableState
+import androidx.compose.runtime.mutableIntStateOf
+import androidx.compose.runtime.mutableStateOf
+import androidx.lifecycle.ViewModel
+import androidx.lifecycle.ViewModelProvider
+import androidx.lifecycle.viewmodel.CreationExtras
+import androidx.preference.PreferenceManager
+import com.google.firebase.analytics.FirebaseAnalytics
+import com.google.firebase.analytics.ktx.analytics
+import com.google.firebase.analytics.ktx.logEvent
+import com.google.firebase.ktx.Firebase
+import com.google.firebase.quickstart.analytics.kotlin.data.Constants
+
+class FirebaseAnalyticsViewModel (
+    private val firebaseAnalytics: FirebaseAnalytics
+): ViewModel() {
+
+    val showFavoriteFoodDialog = mutableStateOf(false)
+
+    private val _selectedImageIndex = mutableIntStateOf(0)
+    val selectedImageIndex: MutableState<Int> = _selectedImageIndex
+
+    private val _userFavoriteFood = mutableStateOf<String?>(null)
+    val userFavoriteFood: MutableState<String?> = _userFavoriteFood
+
+    fun setSelectedImageIndex(index: Int) {
+        _selectedImageIndex.intValue = index
+    }
+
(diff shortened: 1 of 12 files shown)


analytics/app/src/main/java/com/google/firebase/quickstart/analytics/kotlin/FirebaseAnalyticsViewModel.kt
```diff
@@ -0,0 +1,83 @@
...
+    val selectedImageIndex: MutableState<Int> = _selectedImageIndex
+
+    private val _userFavoriteFood = mutableStateOf<String?>(null)
+    val userFavoriteFood: MutableState<String?> = _userFavoriteFood
+
+    fun setSelectedImageIndex(index: Int) {
+        _selectedImageIndex.intValue = index
+    }
+
+    fun setUserFavoriteFood(context: Context, food: String?) {
```
thatfiredev, line 33: It's not a good practice to pass a context into a ViewModel.
Instead, I think we should pass a `SharedPreferences` object to the ViewModel's constructor:

```kotlin
class FirebaseAnalyticsViewModel (
    private val firebaseAnalytics: FirebaseAnalytics,
    private val sharedPreferences: SharedPreferences // <-- this is created in the UI layer
): ViewModel() {
    // ...
}
```

And we'd update the ViewModel Factory to allow passing a SharedPreferences object:
```kotlin
    fun Factory(
        firebaseAnalytics: FirebaseAnalytics = Firebase.analytics,
        sharedPreferences: SharedPreferences
    ): ViewModelProvider.Factory = object : ViewModelProvider.Factory {
            @Suppress("UNCHECKED_CAST")
            override fun <T : ViewModel> create(
                modelClass: Class<T>,
                extras: CreationExtras
            ): T {
                return FirebaseAnalyticsViewModel(firebaseAnalytics, sharedPreferences) as T
            }
        }
```

And in the UI layer:

```
private val sharedPrefs = PreferenceManager.getDefaultSharedPreferences(context)
private val factory = FirebaseAnalyticsViewModel.Factory(Firebase.analytics, sharedPrefs)

private val viewModel: FirebaseAnalyticsViewModel by viewModels { factory }
```

analytics/app/src/main/java/com/google/firebase/quickstart/analytics/kotlin/data/Constants.kt
```diff
@@ -0,0 +1,16 @@
+package com.google.firebase.quickstart.analytics.kotlin.data
```
thatfiredev, line 1: I think this sample apple is small enough that we probably don't need this `data` package. Could we move `Constants.kt` and `ImageInfo.kt` to the `kotlin` package (one directory up from `data`)?

thatfiredev (changes requested):

analytics/app/src/main/java/com/google/firebase/quickstart/analytics/kotlin/FirebaseAnalyticsViewModel.kt
```diff
@@ -0,0 +1,83 @@
...
+import com.google.firebase.analytics.ktx.analytics
+import com.google.firebase.analytics.ktx.logEvent
+import com.google.firebase.ktx.Firebase
+import com.google.firebase.quickstart.analytics.kotlin.data.Constants
+
+class FirebaseAnalyticsViewModel (
+    private val firebaseAnalytics: FirebaseAnalytics
+): ViewModel() {
+
+    val showFavoriteFoodDialog = mutableStateOf(false)
```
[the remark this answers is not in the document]
lehcar09: Thanks for this. I just noticed this and this is very careless of me.

Issue #1647, opened 2024-08-05
Latest Firestore Quickstart encounters NoClassDefFoundError: CredentialsOptions$Builder


argzdev:
### Step 1: Describe your environment

  * Android device: Android Emualtor
  * Android OS version: Android 11
  * Google Play Services version: 21.2.0
  * Firebase/Play Services SDK version: 33.1.2
  
#### Steps to reproduce:

  1. Download Quickstart
  2. Run Firestore
  
#### Observed Results:

  * What happened?  This could be a description, `logcat` output, etc.

```
2024-08-05 19:12:58.712  7297-7297  AndroidRuntime          com...gle.firebase.example.fireeats  E  FATAL EXCEPTION: main
                                                                                                    Process: com.google.firebase.example.fireeats, PID: 7297
                                                                                                    java.lang.NoClassDefFoundError: Failed resolution of: Lcom/google/android/gms/auth/api/credentials/CredentialsOptions$Builder;
                                                                                                    	at com.firebase.ui.auth.util.GoogleApiUtils.getCredentialsClient(GoogleApiUtils.java:29)
                                                                                                    	at com.firebase.ui.auth.viewmodel.AuthViewModelBase.onCreate(AuthViewModelBase.java:30)
                                                                                                    	at com.firebase.ui.auth.viewmodel.ViewModelBase.init(ViewModelBase.java:24)
                                                                                                    	at com.firebase.ui.auth.KickoffActivity.onCreate(KickoffActivity.java:38)
                                                                                                    	at android.app.Activity.performCreate(Activity.java:8000)
                                                                                                    	at android.app.Activity.performCreate(Activity.java:7984)
                                                                                                    	at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1309)
                                                                                                    	at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:3422)
                                                                                                    	at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:3601)
                                                                                                    	at android.app.servertransaction.LaunchActivityItem.execute(LaunchActivityItem.java:85)
                                                                                                    	at android.app.servertransaction.TransactionExecutor.executeCallbacks(TransactionExecutor.java:135)
                                                                                                    	at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:95)
                                                                                                    	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2066)
[...]

PR #1655, opened 2024-08-21
Downgrade play-services-auth

lehcar09:
Temp fix on issue #1647

Changes made:
Downgrade `play-services-auth` to `20.7.0`

### DIFF
--- a/firestore/app/build.gradle.kts
+++ b/firestore/app/build.gradle.kts
@@ -63,7 +63,10 @@ dependencies {
     implementation("com.google.firebase:firebase-auth")
 
     // Google Play services
-    implementation("com.google.android.gms:play-services-auth:21.2.0")
+    // implementation("com.google.android.gms:play-services-auth:21.2.0")
+
+    // Temp fix for issue https://github.com/firebase/quickstart-android/issues/1647
+    implementation("com.google.android.gms:play-services-auth:20.7.0")
 
     // FirebaseUI (for authentication)
     implementation("com.firebaseui:firebase-ui-auth:8.0.2")


firestore/app/build.gradle.kts
```diff
@@ -63,7 +63,10 @@ dependencies {
     implementation("com.google.firebase:firebase-auth")
 
     // Google Play services
-    implementation("com.google.android.gms:play-services-auth:21.2.0")
+    // implementation("com.google.android.gms:play-services-auth:21.2.0")
```
thatfiredev, line 66: We can remove this commented implementation and keep the comment about the temporary fix.

firestore/app/build.gradle.kts
```diff
@@ -63,7 +63,10 @@ dependencies {
...
+
+    // Temp fix for issue https://github.com/firebase/quickstart-android/issues/1647
```
thatfiredev, line 68:
```suggestion
    // Pinned to 20.7.0 as a workaround for issue https://github.com/firebase/quickstart-android/issues/1647
```

firestore/app/build.gradle.kts
```diff
@@ -63,7 +63,10 @@ dependencies {
...
+    implementation("com.google.android.gms:play-services-auth:20.7.0")
```
thatfiredev, line 69: @dpebot will attempt to update this dependency back to 21. Can we please add this dependency to our [blocklist](https://github.com/firebase/quickstart-android/blob/b67eba714c7033ad19599786ba1d7fe0f165e0a9/build.gradle.kts#L67)?

thatfiredev (changes requested):
Thanks @lehcar09 !
I have some comments before we can merge

firestore/app/build.gradle.kts
lehcar09: Will add. Just wanted to check, won't this affect other modules that's uses this dependency (i.e auth, functions)?

thatfiredev: @lehcar09 I think none of them use features from 21.x.x , so it should be ok if we keep them "outdated" for now.

lehcar09: Got it. Thanks!

thatfiredev (approved):
Thanks @lehcar09 !

PR #1671, opened 2024-09-25
Add Firebase Data Connect quickstart

thatfiredev:
TODO in follow-up PRs:
- [ ] Add Search feature
- [ ] Let users update and delete reviews
- [ ] Make sure all the errors are correctly shown in the UI
- [ ] Make sure all the tiles are clickable (movies, actors, reviews)
- [ ] Remove all TODO comments
- [ ] Use the latest schema from the Web quickstart (https://github.com/firebase/quickstart-js/tree/master/dataconnect)

Done in this PR:
- [x] Update README.md to remove steps that were only meant for private preview
- [x] Migrate from alpha to beta

---

~Our GitHub Actions CI will fail at the moment because we're not doing the FDC SDK generation. I will fix that in a follow-up PR~ Update: For now I just removed dataconnect from the CI build  - tracking in #1672.


dataconnect/README.md
```diff
@@ -0,0 +1,97 @@
...
+2. Upgrade your project to the Blaze plan. This lets you create a Cloud SQL
+    for PostgreSQL instance.
+
+    > Note: Though you set up billing in your Blaze upgrade, you won't be
+    charged for usage of Firebase Data Connect or the
+    [default Cloud SQL for PostgreSQL configuration](https://firebase.google.com/docs/data-connect/#pricing)
+    during the preview.
+
+3. Navigate to the [Data Connect section](https://console.firebase.google.com/u/0/project/_/dataconnect)
+    of the Firebase console and follow the setup workflow:
```
marinacoelho, line 29:
```suggestion
    of the Firebase console, click on the "Get Started" button and follow the setup workflow:
```

dataconnect/README.md
```diff
@@ -0,0 +1,97 @@
...
+     - Select a location for your Cloud SQL for PostgreSQL database (this sample uses `us-central1`). If you choose a different location, you'll also need to change the `quickstart-android/dataconnect/dataconnect/dataconnect.yaml` file. 
+     - Fill in the following fields:
```
marinacoelho, line 31:
```suggestion
     - Select the option to create a new Cloud SQL instance and fill in the following fields:
```

dataconnect/README.md
```diff
@@ -0,0 +1,97 @@
...
+       - Service ID: `dataconnect`
+       - Cloud SQL Instance: `fdc-sql`
```
marinacoelho, line 33:
```suggestion
       - Cloud SQL Instance ID: `fdc-sql`
```

dataconnect/README.md
```diff
@@ -0,0 +1,97 @@
...
+       - Database name: `fdcdb`
+4. Allow some time for the Cloud SQL instance to be provisioned. After it's provisioned, the instance
+   can be managed in the [Cloud Console](https://console.cloud.google.com/sql).
+
+5. Download and then add the Firebase Android configuration file (`google-services.json`) to your app:
```
marinacoelho, line 38:
```suggestion
5. If you haven’t already, add an Android app to your Firebase project, with the android package name `com.google.firebase.example.dataconnect`. Download and then add the Firebase Android configuration file (`google-services.json`) to your app:
```

dataconnect/README.md
```diff
@@ -0,0 +1,97 @@
...
+### 3. Cloning the repository
+This repository contains the quickstart to get started with the functionalities of Data Connect.
+
+1. Clone this repository to your local machine:
+   ```sh
+   git clone https://github.com/firebase/quickstart-android.git
+   ```
+
+2. (Private Preview only) Checkout the `fdc-quickstart` branch (`git checkout fdc-quickstart`)
+   and open the project in Android Studio.
```
marinacoelho, line 59:
```suggestion
   and open the project in Android Studio. Make sure you have an Android Studio version that supports AGP 8.6 (from version Koala Feature Drop | 2024.1.2 on).
```

dataconnect/app/src/main/java/com/google/firebase/example/dataconnect/feature/actordetail/ActorDetailScreen.kt
```diff
@@ -0,0 +1,147 @@
...
+import com.google.firebase.example.dataconnect.ui.components.Movie
+import com.google.firebase.example.dataconnect.ui.components.MoviesList
+import com.google.firebase.example.dataconnect.ui.components.ToggleButton
+
+@Composable
+fun ActorDetailScreen(
+    actorId: String,
+    actorDetailViewModel: ActorDetailViewModel = viewModel()
+) {
+    actorDetailViewModel.setActorId(actorId)
```
marinacoelho, line 41: Do you need this function to run every time a recomposition occurs? If not, I recommend calling it from a `LaunchedEffect(true)` block (or `LaunchedEffect(Unit)`). If you use one of this blocks, you create an effect that matches the lifecycle of ActorDetailScreen - meaning that if the screen recomposes, the function shouldn't be called again. This was the recommendation by the Android team on Make it So, the documentation is explains it a bit better: https://developer.android.com/develop/ui/compose/side-effects#rememberupdatedstate

dataconnect/app/src/main/java/com/google/firebase/example/dataconnect/feature/genredetail/Navigation.kt
```diff
@@ -0,0 +1,27 @@
+package com.google.firebase.example.dataconnect.feature.genredetail
+
```
marinacoelho, line 2: I'd recommend renaming the Navigation files to match the nomenclature you're using for each screen. For example, this file would become `GenreDetailNavigation`. It was a bit confusing for me when I had more than one opened at the same time.

dataconnect/app/src/main/java/com/google/firebase/example/dataconnect/feature/actordetail/ActorDetailViewModel.kt
```diff
@@ -0,0 +1,73 @@
...
+            try {
+                val user = firebaseAuth.currentUser
+                val actor = moviesConnector.getActorById.execute(
+                    id = UUID.fromString(actorId)
+                ).data.actor
+
+                _uiState.value = if (user == null) {
+                    ActorDetailUIState.Success(actor, isUserSignedIn = false)
+                } else {
+                    val isFavorite = moviesConnector.getIfFavoritedActor.execute(
```
marinacoelho, line 38: I found it slightly complex to read. Suggestion: rename to `favoriteActor`, then on line 46 you can run `isFavorite = favoriteActor != null`. I believe this will improve the readability a bit.

dataconnect/app/src/main/java/com/google/firebase/example/dataconnect/feature/actordetail/ActorDetailScreen.kt
```diff
@@ -0,0 +1,147 @@
...
+@Composable
+fun ActorDetailScreen(
+    uiState: ActorDetailUIState,
+    onMovieClicked: (actorId: String) -> Unit,
+    onFavoriteToggled: (newValue: Boolean) -> Unit
+) {
+    when (uiState) {
+        is ActorDetailUIState.Error -> ErrorCard(uiState.errorMessage)
+
+        ActorDetailUIState.Loading -> LoadingScreen()
```
marinacoelho, line 63:
```suggestion
        is ActorDetailUIState.Loading -> LoadingScreen()
```

dataconnect/app/src/main/java/com/google/firebase/example/dataconnect/feature/genredetail/GenreDetailScreen.kt
```diff
@@ -0,0 +1,74 @@
...
+    val movies by moviesViewModel.uiState.collectAsState()
+    GenreDetailScreen(movies)
+}
+
+@Composable
+fun GenreDetailScreen(
+    uiState: GenreDetailUIState
+) {
+    when (uiState) {
+        GenreDetailUIState.Loading -> LoadingScreen()
```
marinacoelho, line 37:
```suggestion
        is GenreDetailUIState.Loading -> LoadingScreen()
```

dataconnect/app/src/main/java/com/google/firebase/example/dataconnect/feature/actordetail/ActorDetailScreen.kt
```diff
@@ -0,0 +1,147 @@
...
+
+        is ActorDetailUIState.Success -> {
+            val ui = uiState
```
marinacoelho, line 66: Is there a specific reason why you created a different variable here? It seems like you could keep using `uiState` below.

dataconnect/app/src/main/java/com/google/firebase/example/dataconnect/feature/actordetail/ActorDetailViewModel.kt
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    fun toggleFavorite(newValue: Boolean) {
+        viewModelScope.launch {
+            try {
+                if (newValue) {
+                    moviesConnector.addFavoritedActor.execute(UUID.fromString(actorId))
+                } else {
+                    moviesConnector.deleteFavoriteActor.execute(
+                        userId = firebaseAuth.currentUser?.uid ?: "",
```
marinacoelho, line 62: What happens if we try to execute a mutation with a blank String as `uid`?

dataconnect/app/src/main/java/com/google/firebase/example/dataconnect/ui/components/ActorsList.kt
```diff
@@ -0,0 +1,107 @@
...
+import androidx.compose.ui.text.style.TextOverflow
+import androidx.compose.ui.unit.dp
+import coil.compose.AsyncImage
+
+val ACTOR_CARD_SIZE = 64.dp
+
+/**
+ * Used to represent an actor in a list UI
+ */
+data class Actor(
```
marinacoelho, line 32: Have you thought about having a `model` folder to store `data classes` like this one and the `Movie` one? If you have, what drove the decision to keep them with the UI components?

dataconnect/app/src/main/java/com/google/firebase/example/dataconnect/feature/movies/MoviesViewModel.kt
```diff
@@ -0,0 +1,32 @@
...
+
+    init {
+        viewModelScope.launch {
+            try {
+                val top10Movies = moviesConnector.moviesTop10.execute().data.movies
+                val latestMovies = moviesConnector.moviesRecentlyReleased.execute().data.movies
+
+                _uiState.value = MoviesUIState.Success(top10Movies, latestMovies)
+            } catch (e: Exception) {
+                _uiState.value = MoviesUIState.Error(e.localizedMessage)
```
marinacoelho, line 28:
```suggestion
                _uiState.value = MoviesUIState.Error(e.localizedMessage ?: "")
```

marinacoelho: Reason why I'm suggesting this is because this `localizedMessage` might be `null`, and might cause a crash. That said, if you make this change, it would be nice to use a default error message here as well.

dataconnect/app/src/main/java/com/google/firebase/example/dataconnect/feature/profile/ProfileViewModel.kt
```diff
@@ -0,0 +1,102 @@
...
+        authStateListener = object : AuthStateListener {
+            override fun onAuthStateChanged(auth: FirebaseAuth) {
+                val currentUser = auth.currentUser
+                if (currentUser != null) {
+                    displayUser(currentUser.uid)
+                } else {
+                    _uiState.value = ProfileUIState.AuthState
+                }
+            }
+        }
```
marinacoelho, line 39:
```suggestion
    authStateListener = object : AuthStateListener {
            override fun onAuthStateChanged(auth: FirebaseAuth) {
                val currentUser = auth.currentUser
                if (currentUser != null) {
                    displayUser(currentUser.uid)
                } else {
                    _uiState.value = ProfileUIState.AuthState
                }
            }
        }
```

marinacoelho, line 39:
```suggestion
        authStateListener = AuthStateListener { auth ->
            val currentUser = auth.currentUser
            if (currentUser != null) {
                displayUser(currentUser.uid)
            } else {
                _uiState.value = ProfileUIState.AuthState
            }
        }
```

dataconnect/app/src/main/java/com/google/firebase/example/dataconnect/feature/profile/AuthScreen.kt
```diff
@@ -0,0 +1,94 @@
...
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.text.input.PasswordVisualTransformation
+import androidx.compose.ui.unit.dp
+
+@Composable
+fun AuthScreen(
```
marinacoelho, line 22: Do you plan to add more authentication methods in the future? If so, I'd move this screen to a different `Auth` folder and create it's on `ViewModel` and `UiState`.

marinacoelho (changes requested):
The code is looking very good! Just added some suggestions and some questions/points for us to discuss before merging.

dataconnect/app/src/main/java/com/google/firebase/example/dataconnect/ui/components/ActorsList.kt
thatfiredev: Those are only being used to display lists in the UI. They're not used anywhere else (like as a model for the database operations or anything like that)

The reason why is because Data Connect will already generate models for developers to use in queries in mutations (something like `Query1.ActorItem`, `Query2.ActorItem`, etc), so I'd like them to use those.
I only created these (`Actor`) because they make it easier to pass that data to a UI list, regardless of what the database model looks like (could be `Query1.ActorItem` for example).

thatfiredev: The code comment there was supposed to describe that. Do you think we should update it to be more descriptive ?

marinacoelho: I missed the comment, sorry!  But now that I see it, it makes sense - no need to update it.

dataconnect/app/src/main/java/com/google/firebase/example/dataconnect/feature/actordetail/ActorDetailScreen.kt
thatfiredev: I ended up replacing this with a `SavedStateHandler` in the `ViewModel`, which is the recommended best practice documented [here](https://developer.android.com/topic/libraries/architecture/viewmodel/viewmodel-savedstate). :)

dataconnect/app/src/main/java/com/google/firebase/example/dataconnect/feature/genredetail/Navigation.kt
thatfiredev: I have replaced this with the new [Type safe navigation for Compose](https://medium.com/androiddevelopers/type-safe-navigation-for-compose-105325a97657), and now I don't need those `Navigation.kt` files everywhere =D

dataconnect/app/src/main/java/com/google/firebase/example/dataconnect/feature/actordetail/ActorDetailScreen.kt
thatfiredev: No idea what I was thinking. Fixed

dataconnect/app/src/main/java/com/google/firebase/example/dataconnect/feature/actordetail/ActorDetailViewModel.kt
thatfiredev: Thanks! Done

dataconnect/app/src/main/java/com/google/firebase/example/dataconnect/feature/actordetail/ActorDetailViewModel.kt
```diff
@@ -0,0 +1,73 @@
...
+                        actorId = UUID.fromString(actorId)
+                    )
+                }
+                // Re-run the query to fetch the actor details
+                setActorId(actorId)
+            } catch (e: Exception) {
+                _uiState.value = ActorDetailUIState.Error(e.message ?: "")
```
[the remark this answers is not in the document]
thatfiredev: Good one! I have made the Error messages nullable in all UIStates, and added a default error message to strings.xml that is displayed when the message is null.

dataconnect/app/src/main/java/com/google/firebase/example/dataconnect/feature/movies/MoviesViewModel.kt
thatfiredev: Resolved with the comment above about default error messages.

dataconnect/app/src/main/java/com/google/firebase/example/dataconnect/feature/profile/AuthScreen.kt
thatfiredev: No plans at the moment. I kept it on this same package as a way of showing that it is tied to the profile screen.

dataconnect/app/src/main/java/com/google/firebase/example/dataconnect/feature/actordetail/ActorDetailViewModel.kt
thatfiredev: Good point, I have no idea, but I fear that it deletes everything 😂 

I added a check so that this function returns if there's no signed-in user. This matches the behavior on the Web quickstart.
We can update them altogether for a better user experience where a non-signed-in user will see a proper error message saying they're not signed in.

dataconnect/app/src/main/java/com/google/firebase/example/dataconnect/feature/actordetail/ActorDetailScreen.kt
marinacoelho: Great!

dataconnect/app/src/main/java/com/google/firebase/example/dataconnect/feature/genredetail/Navigation.kt
marinacoelho: Perfect!

dataconnect/app/src/main/java/com/google/firebase/example/dataconnect/feature/actordetail/ActorDetailViewModel.kt
marinacoelho: Yes, I think it's a great idea to add feedback to the users in all quick starts. But for now, I believe that returning if there's no signed in user is enough.

dataconnect/app/src/main/java/com/google/firebase/example/dataconnect/feature/actordetail/ActorDetailViewModel.kt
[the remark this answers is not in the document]
marinacoelho: Good approach :)

marinacoelho (approved):

PR #1701, opened 2024-11-07
Update functions used for testing to be v2

rlazo:
Functions v1 have been deprecated for a while, and we should use v2 when writing functions.

### DIFF
--- a/functions/functions/package.json
+++ b/functions/functions/package.json
@@ -2,7 +2,7 @@
   "name": "functions",
   "description": "Cloud Functions for Android Functions Quickstart",
   "engines": {
-    "node": "22"
+    "node": "20"
   },
   "dependencies": {
     "bad-words": "^1.3.1",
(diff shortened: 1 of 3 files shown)


functions/functions/package.json
```diff
@@ -2,13 +2,13 @@
   "name": "functions",
   "description": "Cloud Functions for Android Functions Quickstart",
   "engines": {
-    "node": "10"
+    "node": "22"
```
thatfiredev, line 5: Most Firebase repos use either Node 18 or 20, as we want to keep compatibility for developers in older versions.

Node 18 will reach EoL on April 2025 ([source](https://nodejs.org/en/about/previous-releases)) - should we go with 20 instead?

thatfiredev (changes requested):
Thanks for updating to the modular Admin SDK too :D

Just one minor change needed before we can merge

functions/functions/package.json
rlazo: Sure, can go with that :-)

thatfiredev (approved):
Thanks @rlazo !

PR #2539, opened 2024-11-15
Upgrade firebase-dataconnect version to 16.0.0-beta03 (was 16.0.0-beta02)

dconeybe:
This PR upgrades the `firebase-dataconnect` dependency to version to `16.0.0-beta03` (https://firebase.google.com/support/release-notes/android#data-connect_v16-0-0-beta03) from `16.0.0-beta02`.

This upgrade is a breaking change and requires Data Connect toolkit version 1.7.0 or later, which is available in firebase-tools version 13.25.0 (https://firebase.google.com/support/release-notes/cli#version_13250_-_november_12_2024). The breaking change is that the code generation for `Date` GraphQL fields and variables uses the new `com.google.firebase.dataconnect.LocalDate` class instead of `java.util.Date`. This new class has functions to convert it to `java.time.LocalDate` and `kotlinx.datetime.LocalDate`.

Since `java.time.LocalDate` was added to the Android standard library in API 26, special case must be taken if your application's version has a `minSdkVersion` less than 26. The options to use `java.time.LocalDate` in such an application are:
1. Just use `com.google.firebase.dataconnect.LocalDate`, without converting it to `java.time.LocalDate`.
2. Put code that uses the `java.time` package inside a `if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O)` block.
3. Enable desugaring (https://developer.android.com/studio/write/java8-support) to make the `java.time.LocalDate` class available all the way back to API 21 (the `minSdkVersion` defined in the `firebase-dataconnect` module).

Also note that using `kotlinx.datetime.LocalDate` requires your application to take a dependency on the "kotlinx-datetime" library by adding `implementation("org.jetbrains.kotlinx:kotlinx-datetime:0.6.1")` to the `dependencies` section of your "app" module's `build.gradle` or `build.gradle.kts`. In addition, the `kotlinx-datetime` library uses `java.time` under the hood, so has the same restrictions and requires the same mitigation strategies mentioned above.

### DIFF
--- a/dataconnect/app/src/main/java/com/google/firebase/example/dataconnect/ui/components/ReviewCard.kt
+++ b/dataconnect/app/src/main/java/com/google/firebase/example/dataconnect/ui/components/ReviewCard.kt
@@ -58,7 +58,7 @@ fun ReviewCard(
                 Text(
                     text =
                         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
-                            val dateFormatter = DateTimeFormatter.ofPattern("MMM d, yyyy", Locale.getDefault())
+                            val dateFormatter = DateTimeFormatter.ofPattern("dd MMM, yyyy", Locale.getDefault())
                             date.toJavaLocalDate().format(dateFormatter)
                         } else {
                             val parseableDateString = date.run {
@@ -70,7 +70,7 @@ fun ReviewCard(
                             val dateParser = SimpleDateFormat("y-M-d", Locale.US)
                             val parsedDate = dateParser.parse(parseableDateString) ?:
                               throw Exception("INTERNAL ERROR: unparseable date string: $parseableDateString")
-                            val dateFormatter = SimpleDateFormat("MMM d, yyyy", Locale.getDefault())
+                            val dateFormatter = SimpleDateFormat("dd MMM, yyyy", Locale.getDefault())
                             dateFormatter.format(parsedDate)
                         },
                     style = MaterialTheme.typography.titleMedium
--- a/dataconnect/app/src/main/java/com/google/firebase/example/dataconnect/ui/components/ReviewCard.kt
+++ b/dataconnect/app/src/main/java/com/google/firebase/example/dataconnect/ui/components/ReviewCard.kt
@@ -54,10 +54,19 @@ fun ReviewCard(
                 modifier = Modifier.padding(bottom = 8.dp)
             ) {
                 Text(
-                    text = SimpleDateFormat(
-                        "dd MMM, yyyy",
-                        Locale.getDefault()
-                    ).format(date.toJavaLocalDate()),
+                    text = run {
+                        val parseableDateString = date.run {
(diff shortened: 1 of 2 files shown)


dataconnect/app/src/main/java/com/google/firebase/example/dataconnect/ui/components/ReviewCard.kt
```diff
@@ -55,10 +56,23 @@ fun ReviewCard(
                 modifier = Modifier.padding(bottom = 8.dp)
             ) {
                 Text(
-                    text = SimpleDateFormat(
-                        "dd MMM, yyyy",
-                        Locale.getDefault()
-                    ).format(date),
+                    text =
+                        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+                            val dateFormatter = DateTimeFormatter.ofPattern("MMM d, yyyy", Locale.getDefault())
```
thatfiredev, line 61: The format we were using before was `dd MMM, yyyy`, should we keep that or was this change intentional?

```suggestion
                            val dateFormatter = DateTimeFormatter.ofPattern("dd MMM, yyyy", Locale.getDefault())
```

dataconnect/app/src/main/java/com/google/firebase/example/dataconnect/ui/components/ReviewCard.kt
```diff
@@ -55,10 +56,23 @@ fun ReviewCard(
...
+                            val parseableDateString = date.run {
+                                val year = "$year".padStart(4, '0')
+                                val month = "$month".padStart(2, '0')
+                                val day = "$day".padStart(2, '0')
+                                "$year-$month-$day"
+                            }
+                            val dateParser = SimpleDateFormat("y-M-d", Locale.US)
+                            val parsedDate = dateParser.parse(parseableDateString) ?:
+                              throw Exception("INTERNAL ERROR: unparseable date string: $parseableDateString")
+                            val dateFormatter = SimpleDateFormat("MMM d, yyyy", Locale.getDefault())
```
thatfiredev, line 73: Same comment as above

```suggestion
                            val dateFormatter = SimpleDateFormat("dd MMM, yyyy", Locale.getDefault())
```

dataconnect/app/src/main/java/com/google/firebase/example/dataconnect/ui/components/ReviewCard.kt
dconeybe: Done.

dataconnect/app/src/main/java/com/google/firebase/example/dataconnect/ui/components/ReviewCard.kt
dconeybe: Done.

thatfiredev (approved):
LGTM

Thanks @dconeybe !

PR #2570, opened 2024-11-18
dataconnect: .gitignore updated with some files generated by the data connect toolkit and/or firebase-tools

dconeybe:


dataconnect/.gitignore
```diff
@@ -15,3 +15,7 @@
 local.properties
 .dataconnect/
 .firebaserc
+.firebase/
+dataconnect-debug.log
+firebase-debug.log
+ui-debug.log
```
thatfiredev, line 21: Shall we do `*-debug.log` or even `*.log` ? I can't imagine us adding any `.log` files to the repo. And even if that happens at some point, we could update the .gitignore to allow it.

dconeybe: Sounds good . I'll make that change today.

dconeybe: Done.

thatfiredev (approved):
LGTM

PR #2649, opened 2025-02-06
Add Credential Manager to Auth samples

marinacoelho:
This PR:

- Removes Google One Tap sign-in, which is deprecated 
- Adds Credential Manager, which is the recommended way to fetch users' credentials

### DIFF
--- a/auth/app/src/main/java/com/google/firebase/quickstart/auth/kotlin/GoogleSignInFragment.kt
+++ b/auth/app/src/main/java/com/google/firebase/quickstart/auth/kotlin/GoogleSignInFragment.kt
@@ -115,6 +115,7 @@ class GoogleSignInFragment : BaseFragment() {
                 // Check if credential is of type Google ID
                 if (credential is CustomCredential && credential.type == TYPE_GOOGLE_ID_TOKEN_CREDENTIAL) {
                     val googleIdTokenCredential = GoogleIdTokenCredential.createFrom(credential.data)
+                    // Sign in to Firebase with the Google ID Token
                     firebaseAuthWithGoogle(googleIdTokenCredential.idToken)
                 } else {
                     Log.d(TAG, "Credential is not of type Google ID!")
--- a/auth/app/src/main/java/com/google/firebase/quickstart/auth/kotlin/GoogleSignInFragment.kt
+++ b/auth/app/src/main/java/com/google/firebase/quickstart/auth/kotlin/GoogleSignInFragment.kt
@@ -1,20 +1,22 @@
 package com.google.firebase.quickstart.auth.kotlin
 
-import android.app.PendingIntent
-import android.content.Intent
-import android.content.IntentSender
 import android.os.Bundle
 import android.util.Log
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
-import androidx.activity.result.IntentSenderRequest
-import androidx.activity.result.contract.ActivityResultContracts.StartIntentSenderForResult
-import com.google.android.gms.auth.api.identity.BeginSignInRequest
-import com.google.android.gms.auth.api.identity.GetSignInIntentRequest
-import com.google.android.gms.auth.api.identity.Identity
-import com.google.android.gms.auth.api.identity.SignInClient
-import com.google.android.gms.common.api.ApiException
+import androidx.credentials.ClearCredentialStateRequest
+import androidx.credentials.Credential
+import androidx.credentials.CredentialManager
+import androidx.credentials.CustomCredential
+import androidx.credentials.GetCredentialRequest
+import androidx.credentials.exceptions.ClearCredentialException
+import androidx.credentials.exceptions.GetCredentialException
+import androidx.lifecycle.lifecycleScope
+import com.google.android.libraries.identity.googleid.GetGoogleIdOption
+import com.google.android.libraries.identity.googleid.GetSignInWithGoogleOption
+import com.google.android.libraries.identity.googleid.GoogleIdTokenCredential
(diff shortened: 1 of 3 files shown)


auth/app/src/main/java/com/google/firebase/quickstart/auth/kotlin/GoogleSignInFragment.kt
```diff
@@ -51,21 +49,19 @@ class GoogleSignInFragment : BaseFragment() {
...
         // Initialize Firebase Auth
         auth = Firebase.auth
 
-        // Display One-Tap Sign In if user isn't logged in
-        val currentUser = auth.currentUser
-        if (currentUser == null) {
-            oneTapSignIn()
-        }
+        // Button listeners
+        binding.signInButton.setOnClickListener { signInWithButton() }
```
thatfiredev, line 59: Maybe call this function "signIn()" ? And the bottom one "showBottomSheet()" ?

auth/app/src/main/java/com/google/firebase/quickstart/auth/kotlin/GoogleSignInFragment.kt
```diff
@@ -75,23 +71,58 @@ class GoogleSignInFragment : BaseFragment() {
...
+        val signInWithGoogleOption = GetSignInWithGoogleOption
+            .Builder(serverClientId = requireContext().getString(R.string.default_web_client_id))
+            .build()
+
+        // Create the Credential Manager request using the configuration created above
+        val request = GetCredentialRequest.Builder()
+            .addCredentialOption(signInWithGoogleOption)
+            .build()
+
+        getCredential(request)
```
thatfiredev, line 85: maybe name this "fetchCredential", since the `get` prefix is often used for getters in Java/Kotlin

auth/app/src/main/java/com/google/firebase/quickstart/auth/kotlin/GoogleSignInFragment.kt
```diff
@@ -75,23 +71,58 @@ class GoogleSignInFragment : BaseFragment() {
...
+    }
+
+    private fun getCredential(request: GetCredentialRequest) {
+        viewLifecycleOwner.lifecycleScope.launch {
+            try {
+                // Launch Credential Manager UI
+                val result = credentialManager.getCredential(
+                    request = request,
+                    context = requireContext()
```
thatfiredev, line 109: In Android it's more common to pass the `Context` first

```suggestion
                    context = requireContext(),
                    request = request                    
```

auth/app/src/main/java/com/google/firebase/quickstart/auth/kotlin/GoogleSignInFragment.kt
```diff
@@ -75,23 +71,58 @@ class GoogleSignInFragment : BaseFragment() {
...
+                // Extract credential from the result returned by Credential Manager
+                val credential = result.credential
+
+                // Check if credential is of type Google ID
+                if (credential is CustomCredential && credential.type == TYPE_GOOGLE_ID_TOKEN_CREDENTIAL) {
+                    val googleIdTokenCredential = GoogleIdTokenCredential.createFrom(credential.data)
+                    // Sign in to Firebase with the Google ID Token
+                    firebaseAuthWithGoogle(googleIdTokenCredential.idToken)
+                } else {
+                    Log.d(TAG, "Credential is not of type Google ID!")
```
thatfiredev, line 121: Maybe log with a warning instead (`Log.w`) ?

auth/app/src/main/java/com/google/firebase/quickstart/auth/kotlin/GoogleSignInFragment.kt
```diff
@@ -117,60 +148,20 @@ class GoogleSignInFragment : BaseFragment() {
...
-
     private fun signOut() {
         // Firebase sign out
         auth.signOut()
 
-        // Google sign out
-        signInClient.signOut().addOnCompleteListener(requireActivity()) {
-            updateUI(null)
+        // When a user signs out, clear the current user credential state from all credential providers.
+        // This will notify all providers that any stored credential session for the given app should be cleared.
```
thatfiredev, line 156: Should this be done when the user signs out or when the user deletes their account?

I'm thinking of a situation when a user signs out from Account A so that they can sign in with Account B.
But if they ever want to sign in with Account A again, will the Bottom sheet still show up? Or will it have been cleared?

marinacoelho: Yes - it will! This is just to avoid limiting sign-in options for future get-credential calls, since some providers may prioritize the active credential over any other available credential. According to the [documentation](https://developer.android.com/identity/sign-in/credential-manager-siwg#handle-sign-out): "When your user explicitly signs out of your app and in order to get the holistic sign-in options the next time, you should call this API to let the provider clear any stored credential session."

auth/app/src/main/java/com/google/firebase/quickstart/auth/kotlin/GoogleSignInFragment.kt
marinacoelho: Done!

auth/app/src/main/java/com/google/firebase/quickstart/auth/kotlin/GoogleSignInFragment.kt
marinacoelho: Changed to `createGoogleIdToken`

thatfiredev (approved):

PR #2670, opened 2025-05-22
[firebase-ai] Add missing entries to libs.versions.toml

rlazo:
Some of the versions were declared in the top-level `libs.versions.toml` file instead.

Also, sorted alphabetically the entries in the file.


thatfiredev (approved):
LGTM

PR #2667, opened 2025-05-19
feat: create Firebase AI quickstart

thatfiredev:
CI will fail at the moment, since ~the SDK has not been released yet~. I will make sure it's green after the SDK is released and only merge by then.

Fast follows:
- [ ] Add a Function calling example
- [ ] Add a Stream Realtime example

PR #2678, opened 2025-06-17
feat(firebase-ai): add function calling example

thatfiredev:
This is a follow up to #2667


firebase-ai/app/src/main/java/com/google/firebase/quickstart/ai/feature/text/functioncalling/WeatherRepository.kt
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * Hypothetical repository that calls an external weather API.
+ */
+class WeatherRepository {
+
+    companion object {
+        suspend fun fetchWeather(
+            city: String, state: String, date: String
+        ): JsonObject = withContext(Dispatchers.IO) {
+            // For demo purposes, this hypothetical response is
```
marinacoelho, line 17: Same as above, I'd rephrase it to something along the lines of `// This response is a hardcoded example of the expected format, and should be used for demo purposes only`

thatfiredev: Marina and I discussed offline and we might end up changing the docs instead.

I'll merge this as it is right now - we can change it once the docs are updated.

PR #2682, opened 2025-07-02
feat(fireabase-ai): add keyboard focus management

YoungHypo:
The soft keyboard is not automatically retracted when the user clicks on other areas of the interface after typing in the text input box.

By adding focus management logic to the root containing the text input box, the focus is automatically cleared when the user clicks on a blank area, which triggers a keyboard retraction.


thatfiredev:
@YoungHypo Thanks for putting together a PR.

I think the current behavior is intended. The keyboard should be dismissed by clicking on the back button, not when clicking on a blank area.

I have noticed this same behavior in popular chat apps such as WhatsApp, Telegram and Google Chat.

That said, I'm going to close this PR for now, but please leave a comment if you'd like to discuss more.

PR #2685, opened 2025-07-04
Update `README.md` to reference firebase ai logic sample app

dlarocque:
Update `README.md` to reference the new Firebase AI Logic sample app rather than the old Vertex AI sample app.


thatfiredev (approved):
Thanks @dlarocque !

PR #2686, opened 2025-07-04
chore(ai): use kotlin 2.2.0

thatfiredev:
It seems like dpebot missed this in https://github.com/firebase/quickstart-android/pull/2681

The reason why it didn't update the build.gradle.kts file is because that file used the `kotlin(...)` function instead of `id(...)`.

I need to investigate why it didn't update the TOML as well (see https://github.com/firebase/quickstart-android/pull/2681#issuecomment-3036667136)

### DIFF
--- a/firebase-ai/app/build.gradle.kts
+++ b/firebase-ai/app/build.gradle.kts
@@ -2,7 +2,7 @@ plugins {
     id("com.android.application")
     id("org.jetbrains.kotlin.android")
     id("org.jetbrains.kotlin.plugin.compose")
-    kotlin("plugin.serialization") version "2.1.21"
+    id("org.jetbrains.kotlin.plugin.serialization")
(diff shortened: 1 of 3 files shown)


sadnanshawon:
val apiKey = BuildConfig.GEMINI_API_KEY
val client = OkHttpClient()

val request = Request.Builder()
    .url("https://generativelanguage.googleapis.com/v1beta/models/gemini-pro:generateContent?key=$apiKey")
    .post(
        RequestBody.create(
            MediaType.parse("application/json"),
            """{"contents":[{"parts":[{"text":"Hello Gemini"}]}]}"""
        )
    )
    .build()

val response = client.newCall(request).execute()
val responseBody = response.body()?.string()
Log.d("GeminiResponse", responseBody ?: "No response")

PR #2767, opened 2026-03-06
refactor(ai-logic): split samples into various ViewModels

thatfiredev:
The goal is to make these samples more readable by splitting them into various ViewModels - each sample/feature has its own ViewModel.

There should be no UI change (other than typo fixes) or functionality changes.


rlazo (approved):
Few comments, but nothing blocking .

PR #2791, opened 2026-05-19
refactor(ai-logic): simplify samples categories

thatfiredev:
This PR should:
- Simplify the categories we have: Gemini 3, Nano Banana, Multimodal understanding, Tools and function calling, Live API streaming, Server prompt templates and Hybrid inference.
- Introduce new samples for Nano Banana 2 (preview) and Nano Banana Pro (preview).
- Update the SVG generator to use Gemini 3.5 Flash with Thinking Level High.
- Remove the Gemini 2.5 Flash example
- Remove the "Travel tips" and "Chatbot recommendations for courses" samples

### Preview

result


firebase-ai/app/src/main/java/com/google/firebase/quickstart/ai/ui/navigation/MainMenuScreen.kt
```diff
@@ -77,21 +75,16 @@ fun MenuScreen(
                     } else {
                         null
                     },
-                    modifier = Modifier.padding(end = 8.dp)
+                    modifier = Modifier.padding(horizontal = 4.dp)
                 )
             }
         }
-        Text(
-            text = "Samples",
```
marinacoelho, line 85: Do we keep Strings hardcoded in the screen for the quick start samples? If not - recommend moving all hardcoded strings in the compose files to the xml file.

marinacoelho (approved):
Just one comment about strings - LGTM.

firebase-ai/app/src/main/java/com/google/firebase/quickstart/ai/ui/navigation/MainMenuScreen.kt
thatfiredev: I don't have a strong preference.
Btw, this line was deleted, so we technically need this particular string.

PR #2804, opened 2026-06-04
refactor(firebase-ai): use explicit backing fields and the new onDeviceExtension

thatfiredev:
- Kotlin 2.4.0 marks [explicit backing fields](https://kotlinlang.org/docs/properties.html#explicit-backing-fields) as stable
- Use the new onDeviceExtension for hybrid inference


thatfiredev:
Ignoring the Gemini reviews because the code compiles successfully

marinacoelho (approved):

